PR #55, opened 2020-11-24
add cmdargs

### LINKED ISSUE #18: Pass through command line arguments
In tox you can do `tox -e py36-redis -- "-k only_test_this_one"`. This passes `"-k only_test_this_one"` as a variable called `posargs` in tox which allows you to specify it in your test command. We'd then typically have something like `redis: pytest {posargs} tests/contrib/redis` which would use it.

For Riot, I don't think we need to do the whole `posargs` thing, but we definitely want to be able to pass through arguments to the underlying test command.

Maybe something like

```
riot redis --args="-k 'only_test_this_one'"
```

majorgreys:
Resolves #18


brettlangdon:
Click supports extra arguments via `riot run suite -- --extra -args=5`


```python
import click

@click.command()
@click.option("-t", "--test", default=False, is_flag=True)
@click.pass_context
def main(ctx, test):
   print(f"Test: {test}")
   print(f"Extra args: {ctx.args}")

main.allow_extra_args = True  # I haven't found a better way to configure this


if __name__ == "__main__":
    main()
```


```
$ python main.py
Test: False
Extra args: []

$ python main.py --test
Test: True
Extra args: []

$ python main.py -- --test -args=5
Test: False
Extra args: ['--test', '-args=5']

$ python main.py --test -- --test -args=5
Test: True
Extra args: ['--test', '-args=5']
```

brettlangdon:
https://click.palletsprojects.com/en/7.x/api/#click.BaseCommand.allow_extra_args

majorgreys:
I had seen that but went with the more explicit approach to setting an infix `cmdargs` argument. In addition to breaking a convention devs might be used to, a drawback to this more explicit approach is that what follows `--cmdargs` has to be wrapped with quotes. I will start a PR for this change to get comments.

brettlangdon:
👍🏻 yeah, I agree with your assessment, think the passing all args in a quoted string is an anti-pattern these days.

PR #57, opened 2020-11-24
colorizing output

### LINKED ISSUE #46: Colorize end test result
`tox` has this great feature where the end of the log is green or red depending on the output.
The current check or cross marks are great, but they miss colors to stand out.

Isabella-Pham:
closes #46


riot/riot.py
```diff
@@ -213,13 +214,48 @@ def run_suites(
             finally:
                 results.append(result)
 
-        print("\n-------------------summary-------------------", file=out)
+        def is_warning(output):
```
jd, line 217:
```suggestion
       @staticmethod
        def is_warning(output):
```

riot/riot.py
```diff
@@ -213,13 +214,48 @@ def run_suites(
...
+                "deprecation", 
+                "warning", 
+                "no longer maintained", 
+                "not maintained",
+                "did you mean",
+                "syntaxwarning",
+                "userwarning",
+                "runtimewarning"
+            ]
+            for warning in warnings:
```
jd, line 229: maybe set this as a class variable so it's easy to override

riot/riot.py
```diff
@@ -213,13 +214,48 @@ def run_suites(
...
+                if warning in lower_output:
+                    return True
+            return False
```
jd, line 232:
```suggestion
            lower_output = output.lower()
            return any(lower_output in warning for warning in ("deprecated", 
                "deprecation", 
                "warning", 
                "no longer maintained", 
                "not maintained",
                "did you mean",
                "syntaxwarning",
                "userwarning",
                "runtimewarning"
            ))
```
;)

riot/riot.py
```diff
@@ -213,13 +214,48 @@ def run_suites(
...
+
+        click.echo(click.style("\n-------------------summary-------------------", bold=True))
```
jd, line 234: haha back to tox style 😁

tests/test_riot.py
```diff
@@ -106,3 +107,14 @@ def test_suites_iter():
 
     instances = list(suites_iter(suites, pattern=re.compile(".*")))
     assert len(instances) == (2 * 3 * 2) + (2 * 1 * 2) + (2 * 3 * 4)
+
+def test_output():
```
jd, line 111:
```suggestion
def test_output() -> None:
```

jd (changes requested):

riot/riot.py
jd: I realize my original comment was weird because the method is actually a closure.
There's no need to define a closure here (i.e. to re-define the function on every execution of this method), so you should move it outside the method as a method :)

Kyle-Verhoog: @Isabella-Pham this still needs to be addressed 🙂

tests/test_output_display.py
```diff
@@ -0,0 +1,13 @@
+import warnings
```
Kyle-Verhoog, line 1: This file isn't being used anymore

Kyle-Verhoog (approved):

Issue #33, opened 2020-10-19
Add user documentation


Kyle-Verhoog:
Once we're stable we should do the whole sphinx + readthedocs kinda deal. Updating the README would also be nice.

PR #60, opened 2020-11-24
Add sphinx docs and readthedocs config

Kyle-Verhoog:
Clears a path for #33


docs/Makefile
```diff
@@ -0,0 +1,12 @@
+SPHINXOPTS    ?=
```
jd, line 1: don't really need this file…

jd (changes requested):
just remove the Makefile

docs/Makefile
Kyle-Verhoog: lol true

jd (approved):

PR #62, opened 2020-11-24
Add support for singleton values

Kyle-Verhoog:
This allows pkgs, env and pys to have singleton values which makes the
configuration more ergonomic.


riot/riot.py
```diff
@@ -24,14 +23,24 @@ def __init__(self, *args, **kwargs):
         self.__dict__ = self
 
 
+def rm_singletons(
+    d: t.Dict[str, t.Union[str, t.List[str]]]
+) -> t.Dict[str, t.List[str]]:
+    return {k: [v] if isinstance(v, str) else v for k, v in d.items()}
```
jd, line 29:
```suggestion
_K = t.TypeVar("_K")

def rm_singletons(
    d: t.Dict[_K, t.Union[str, t.List[str]]]
) -> t.Dict[_K, t.List[str]]:
    return {k: [v] if isinstance(v, str) else v for k, v in d.items()}
```

riot/riot.py
```diff
@@ -24,14 +23,24 @@ def __init__(self, *args, **kwargs):
...
+
+
+def float_to_list(x: t.Union[float, t.List[float]]) -> t.List[float]:
+    return [x] if isinstance(x, float) else x
```
jd, line 33:
```suggestion
_K = t.TypeVar("_K")
def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
    return [x] if not isinstance(x, collections.abc.Iterable) else x
```

and use it in `rm_singletons`

tests/test_cli.py
```diff
@@ -385,3 +385,54 @@ def test_failure():
         result = cli.invoke(riot.cli.main, ["run", "-s", "failure"])
         assert result.exit_code == 1
         assert result.stdout == ""
+
+
+def test_types(cli: click.testing.CliRunner):
```
jd, line 390:
```suggestion
def test_types(cli: click.testing.CliRunner) -> None:
```

jd (changes requested):

riot/riot.py
Kyle-Verhoog: Hmm I had tried something like this initially and had gotten:

image

I wasn't using `collections.abc.Iterable` though, so maybe that was it!

Kyle-Verhoog:
Yeah @jd getting the same type error as I got before:


image


Not quite sure how to resolve 🤔

riot/riot.py
```diff
@@ -24,14 +23,27 @@ def __init__(self, *args, **kwargs):
...
 
+_K = t.TypeVar("_K")
+
+def rm_singletons(
+    d: t.Dict[_K, t.Union[str, t.List[str]]]
+) -> t.Dict[_K, t.List[str]]:
+    return {k: [v] if isinstance(v, str) else v for k, v in d.items()}
+
+
+_K = t.TypeVar("_K")
```
jd, line 34: You don't need to redefine it here, you can use the same one.

riot/riot.py
```diff
@@ -32,15 +31,29 @@ def __init__(self, *args, **kwargs):
         self.__dict__ = self
 
 
+def rm_singletons(d: t.Dict[_K, t.Union[str, t.List[str]]]) -> t.Dict[_K, t.List[str]]:
+    return {k: [v] if isinstance(v, str) else v for k, v in d.items()}
```
jd, line 35:
```suggestion
def rm_singletons(d: t.Dict[_K, t.Union[_V, t.List[_V]]]) -> t.Dict[_K, t.List[_V]]:
    return {k: to_list(v) for k, v in d.items()}
```

jd (changes requested):

riot/riot.py
```diff
@@ -32,15 +31,29 @@ def __init__(self, *args, **kwargs):
         self.__dict__ = self
 
 
+def rm_singletons(d: t.Dict[_K, t.Union[_V, t.List[_V]]]) -> t.Dict[_K, t.List[_V]]:
+    return {k: to_list(v) for k, v in d.items()}
+
+
+def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
+    return [x] if not isinstance(x, list) else x
```
jd, line 39: I'm approving this but having 2 simple unit tests for those functions could be worth the 5 minutes.
Feel free to add them later.

jd (approved):

riot/riot.py
Kyle-Verhoog: seems like a good place for some doctests!

Kyle-Verhoog: Done in https://github.com/DataDog/riot/pull/77

PR #63, opened 2020-11-24
chore(mypy): tweak mypy options and add more typing

jd:
- Enable `ignore_missing_imports` so we don't fail on missing typing for
  external libs
- Disallow incomplete defs to make sure we have fully documented type on every
  typed function
- Warn on unused configs and ignore directives
- Disallow usage of Any
- Specify files to check


brettlangdon (approved):

brettlangdon (approved):

Kyle-Verhoog (approved):
🙌

PR #67, opened 2020-11-24
refactor: replace attrs by dataclasses

jd:
We don't really need/use any fancy feature for attrs.


jd:
Damn I missed it did not work on Python 3.6. Do we care about 3.6? 🤔 😆

Kyle-Verhoog:
We could use the backport only for 3.6: https://pypi.org/project/dataclasses/

Kyle-Verhoog (approved):

PR #69, opened 2020-11-24
fix: do not specify encoding

jd:
riot should just be a serving hatch and not mess with user encoding in any way,
use the system default one.

### DIFF
--- a/riot/riot.py
+++ b/riot/riot.py
@@ -14,7 +14,7 @@ logger = logging.getLogger(__name__)
 
 
 SHELL = "/bin/bash"
-ENCODING = "utf-8"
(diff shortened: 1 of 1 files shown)


Kyle-Verhoog (approved):

Kyle-Verhoog:
nice

brettlangdon:
Does this have any impact on the unicode characters we use for printing
status?

jd:
> Does this have any impact on the unicode characters we use for printing status?

No it does not.

For that I think we have to hope the user is using a terminal with UTF-8 encoding. We could detect it I guess… I'll wait for someone to complain.

Kyle-Verhoog (approved):

PR #70, opened 2020-11-24
Drop support for Python 3.6

Kyle-Verhoog:
If we so decide to do so 😄


brettlangdon (approved):

jd (approved):

jd:
Whoever is admin needs to fix branch protection

PR #71, opened 2020-11-24
fix: do not use cp to copy trees

jd:
The current use of `cp` is more-or-less portable but is broken:

`cp -r foo bar` copies foo as bar the first time but `cp -r foo bar` copies
`foo` *into* `bar` the second time.

### DIFF
--- a/riot/riot.py
+++ b/riot/riot.py
@@ -192,7 +192,7 @@ class Session:
                     )
                     try:
                         shutil.copytree(base_venv, venv_name)
-                    except FileNotFoundError as e:
+                    except FileNotFoundError:
                         logger.info("Base virtualenv '%s' does not exists", venv_name)
                         continue
                     except FileExistsError:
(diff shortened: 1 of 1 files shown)


Kyle-Verhoog (dismissed):
great catch!

riot/riot.py
```diff
@@ -191,14 +191,13 @@ def run(
...
-                        run_cmd(
-                            ["cp", "-r", base_venv, venv_name], stdout=subprocess.PIPE
-                        )
-                    except CmdFailure as e:
-                        raise CmdFailure(
-                            f"Failed to create virtualenv '{venv_name}'\n{e.proc.stdout}",
-                            e.proc,
-                        )
+                        shutil.copytree(base_venv, venv_name)
+                    except FileNotFoundError as e:
```
Kyle-Verhoog, line 195:
```suggestion
                    except FileNotFoundError:
```

riot/riot.py
```diff
@@ -191,14 +191,13 @@ def run(
...
+                        logger.info("Base virtualenv '%s' does not exists", venv_name)
```
Kyle-Verhoog, line 196:
```suggestion
                        logger.info("Base virtualenv '%s' does not exist", venv_name)
```

Kyle-Verhoog (approved):

PR #79, opened 2020-11-24
Support multiple interpreter definitions

### LINKED ISSUE #30: Allow specification of python executable
Right now riot assumes that eg. `python2.7` , `python3.7` executables exist. If a user has these executables named differently or don't have them on the path then they cannot use riot.

Kyle-Verhoog:
Adds support for specifying arbitrary interpreters to be used with riot.

```
Venv(
  pys=[3, 3.7, "pypy3", "/path/to/my/interp"]
)
```

The way this works is by first checking the PATH for the raw value given (eg. `3.7`) and then falling back to checking the PATH for `python{raw value}` (eg. `python3.7`).

Fixes #30

### DIFF
--- a/riot/riot.py
+++ b/riot/riot.py
@@ -221,6 +221,7 @@ class Session:
     def run(
         self,
         pattern: t.Pattern[str],
+        venv_pattern: t.Pattern[str],
         skip_base_install: bool = False,
         recreate_venvs: bool = False,
         out: t.TextIO = sys.stdout,
@@ -252,17 +253,19 @@ class Session:
             }
 
             if pkgs:
-                # Strip special characters for the venv directory name.
-                venv_postfix = "_".join(
-                    [f"{n}{rmchars('<=>.,', v)}" for n, v in pkgs.items()]
-                )
-                venv_name = f"{base_venv}_{venv_postfix}"
+                venv_name = get_venv_directory_name(base_venv, pkgs)
                 pkg_str = " ".join(
                     [f"'{get_pep_dep(lib, version)}'" for lib, version in pkgs.items()]
                 )
             else:
                 venv_name = base_venv
                 pkg_str = ""
+
+            if not venv_pattern.search(venv_name):
+                logger.debug(
+                    "Skipping venv instance '%s' due to pattern mismatch", venv_name
+                )
+                continue
 
             # Result which will be updated with the test outcome.
             result = VenvInstanceResult(
@@ -346,9 +349,14 @@ class Session:
         if any(True for r in results if r.code != 0):
             sys.exit(1)
 
-    def list_venvs(self, pattern, pythons=None, out=sys.stdout):
+    def list_venvs(self, pattern, venv_pattern, pythons=None, out=sys.stdout):
         for inst in self.venv.instances(pattern=pattern):
             if pythons and inst.py not in pythons:
+                continue
+            base_venv = get_base_venv_path(inst.py)
+            if not venv_pattern.search(
+                get_venv_directory_name(base_venv, dict(inst.pkgs))
+            ):
                 continue
             pkgs_str = " ".join(
                 f"'{get_pep_dep(name, version)}'" for name, version in inst.pkgs
@@ -411,6 +419,12 @@ def get_pep_dep(libname: str, version: str) -> str:
     ref: https://www.python.org/dev/peps/pep-0508/
     """
     return f"{libname}{version}"
+
+
+def get_venv_directory_name(base_venv, pkgs):
+    # Strip special characters for the venv directory name.
+    venv_postfix = "_".join([f"{n}{rmchars('<=>.,', v)}" for n, v in pkgs.items()])
+    return f"{base_venv}_{venv_postfix}"
 
 
 def get_env_str(envs: t.Sequence[t.Tuple[str, str]]) -> str:
--- a/riot/riot.py
+++ b/riot/riot.py
@@ -61,9 +61,82 @@ def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
     return [x] if not isinstance(x, list) else x
 
 
+class Interpreter:
+    def __init__(self, hint: t.Union[str, float, int]):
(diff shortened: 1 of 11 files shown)


riot/riot.py
```diff
@@ -393,6 +393,9 @@ def create_base_venv(pyversion, path=None, recreate=True):
     py_ex = f"python{pyversion}"
     py_ex = shutil.which(py_ex)
 
+    if not py_ex:
+        py_ex = shutil.which(pyversion):
```
brettlangdon, line 397: I am not so sure about this.... because if you are asking for `python 3.8` and it isn't found, we are just going to use `python 2.7` because it is available?

Kyle-Verhoog: So all this does is if like `python3.8` isn't available then it's going to check for `3.8`.

Presumably `pythonmypy` isn't going to exist so then we'd try `mypy`. Doing the check in the opposite order probably makes more sense though.

brettlangdon:
```python
pys=3.8
pys=[3.8]
pys=["pypy", 3.8]
pys=["pypy3", 3.8]
pys="pypy3"
pys=["python3.9"]
pys=["/path/to/some/custom/executable"]
```

e.g. if only a number is given, it is assumed `f"python{py}"`, otherwise, it is assumed the name of the executable was given?

brettlangdon:
as long as the executable we resolve works with `virtualenv` then should be good

brettlangdon:
oh right, I read this wrong, or read the title/description wrong

brettlangdon:
Another idea is to just resolve float(3.8) to "python3.8" as soon as
possible, then "pys"/"pythons" being passed around everywhere is just a str
list of python executable names.

riot/riot.py
```diff
@@ -61,9 +61,82 @@ def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
...
+        False
+        >>> Interpreter("3.8") == Interpreter(3.8)
+        True
+        """
+        if isinstance(other, Interpreter):
+            return other._hint == self._hint
+        else:
+            return False
+
+    def __hash__(self):
```
jd, line 85: You could leverage `dataclasses` to have this I think.

riot/riot.py
```diff
@@ -61,9 +61,82 @@ def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
...
+
+    def __repr__(self) -> str:
+        """Return the representation of the Interpreter.
+
+        >>> repr(Interpreter(3.9)) == repr(Interpreter(3.9))
+        True
+        >>> repr(Interpreter("3.9")) == repr(Interpreter(3.9))
+        True
+        """
+        return f"Interpreter({repr(self._hint)})"
```
jd, line 107: ditto

jd (changes requested):
I feel like there's a few missing tests if you implement `__hash__` yourself for example.

riot/riot.py
```diff
@@ -61,9 +61,62 @@ def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
     return [x] if not isinstance(x, list) else x
 
 
+@dataclasses.dataclass(unsafe_hash=True)
+class Interpreter:
+    hint: dataclasses.InitVar[str]
+
+    def __post_init__(self, hint: t.Union[float, int, str]) -> None:
```
jd, line 68: This does not work the way you think:
```python
import dataclasses
import typing as t


@dataclasses.dataclass
class I:
    hint: dataclasses.InitVar[str]

    def __post_init__(self, hint: t.Union[float, int, str]) -> None:
        self.hint = str(hint)
```

```
$ mypy
td.py:12: error: Argument 1 to "I" has incompatible type "int"; expected "str"
```

You want to do this instead:
```python
import dataclasses
import typing as t


@dataclasses.dataclass
class I:
    _T_hint = t.Union[float, int, str]

    hint: dataclasses.InitVar[_T_hint]
    _hint: str = dataclasses.field(init=False)

    def __post_init__(self, hint: _T_hint) -> None:
        self._hint = str(hint)
```

riot/riot.py
```diff
@@ -61,9 +61,62 @@ def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
...
+        """Normalize the data."""
+        self.hint = str(hint)
+
+    def __eq__(self, other: object) -> bool:
```
jd, line 72: This should also be done with dataclass AFAICT, you can use `compare=True` or `False` to configure it.

riot/riot.py
```diff
@@ -61,9 +61,62 @@ def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
...
+        interpreters to be installed for their usage.
+        """
+        py_ex = shutil.which(self.hint)
+        if not py_ex:
+            py_ex = shutil.which(f"python{self.hint}")
+
+        if not py_ex:
+            raise FileNotFoundError(f"Python interpreter {self.hint} not found")
+        else:
+            return py_ex
```
jd, line 114:
```suggestion
        if py_ex:
            return py_ex

         raise FileNotFoundError(f"Python interpreter {self.hint} not found")
```
nitpick 😁

jd (changes requested):

riot/riot.py
```diff
@@ -61,9 +61,49 @@ def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
...
 
 
+@dataclasses.dataclass(unsafe_hash=True, eq=True)
+class Interpreter:
+    _T_hint = t.Union[float, int, str]
+
+    hint: dataclasses.InitVar[_T_hint]
+    _hint: str = dataclasses.field(init=False)
+
+    def __post_init__(self, hint: t.Union[float, int, str]) -> None:
```
jd, line 71:
```suggestion
    def __post_init__(self, hint: _T_hint) -> None:
```

jd (changes requested):
maybe add a few unit tests if you to make sure your hash/eq methods still works as expected?

majorgreys (approved):

PR #80, opened 2020-11-24
Add flake8-import-order

majorgreys:


Kyle-Verhoog:
@majorgreys is there an autoformatter for these? I vaguely recall you mentioning one before.

Kyle-Verhoog (approved):

majorgreys:
> @majorgreys is there an autoformatter for these? I vaguely recall you mentioning one before.

Right, https://pypi.org/project/isort/ would give us an automatic way to check imports. Being we have a small codebase at this point, I opted to first try the `flake8-import-order`.

Kyle-Verhoog:
cool, makes sense to me!

jd:
Yeah I was going to propose `isort`, it's easier to use :)

PR #85, opened 2020-11-25
Provide latest export

### LINKED ISSUE #50: Change syntax for latest version of package

Kyle-Verhoog:
This removes the obscure syntax used for picking the latest package.
"latest" isn't the greatest name since this just corresponds to not
specifying a version. It could be taken from the pip cache and not be
the absolute latest.

Fixes #50


majorgreys:
Being a constant, we should use all caps.

jd (approved):

PR #87, opened 2020-11-25
Add user documentation

### LINKED ISSUE #33: Add user documentation

Kyle-Verhoog:
A follow up PR will be needed to document each of the cli arguments and options but otherwise this should be a pretty solid starting point.

Resolves #33.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -10,11 +10,17 @@
 # import sys
 # sys.path.insert(0, os.path.abspath('.'))
 
-
 project = "riot"
 
+copyright = "2020-Present Datadog, Inc"  # noqa
+
 extensions = [
-    "sphinx.ext.doctest",
     "sphinx.ext.autodoc",
+    "sphinx.ext.doctest",
+    "sphinx.ext.napoleon",
+    "sphinx_rtd_theme",
     "reno.sphinxext",
+    "sphinx_click",
 ]
+
+html_theme = "sphinx_rtd_theme"
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -1,0 +1,70 @@
+Quickstart
+==========
+
+riot is configured with a single Python file typically placed at the root of
+your project and named ``riotfile.py``.
+
+
+Here is a ``riotfile.py`` which defines 5 virtual environment instances. One
+defines how to run ``mypy``, another for ``black`` and 3 instances to run tests
+with ``pytest``::
+
+        from riot import latest, Venv
+
+        version = "0.3"
+
+        venv = Venv(
+            pys=[3.9],
+            venvs=[
+                Venv(
+                    name="fmt",
+                    command="black .",
+                    pkgs={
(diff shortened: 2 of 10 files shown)


docs/conf.py
```diff
@@ -10,11 +10,17 @@
...
+copyright = "2020-Present Datadog, Inc"  # noqa
+
 extensions = [
-    "sphinx.ext.doctest",
     "sphinx.ext.autodoc",
+    "sphinx.ext.doctest",
+    "sphinx.ext.napoleon",
+    "sphinx_rtd_theme",
     "reno.sphinxext",
+    "sphinx_click",
```
jd, line 23: do you need to this something to `docs/requirements.txt`?

docs/quickstart.rst
```diff
@@ -0,0 +1,68 @@
...
+        venv = Venv(
+            pys=[3.9],
+            venvs=[
+                Venv(
+                    name="fmt",
+                    command="black .",
+                    pkgs={
+                        "black": "==20.8b1",
+                    },
+                ),
```
jd, line 23: Reading this I'm wondering what happens if you do:

```python
venv = Venv(
            pys=[3.9],
            venvs=[
                Venv(
                    name="fmt",
                    pys=[3.8],
                    command="black .",
                    pkgs={
                        "black": "==20.8b1",
                    },
                ),
            ],
         )
```

with `pys` specified twice and in contradiction?

riot/riot.py
```diff
@@ -63,6 +63,43 @@ def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
...
+              ),
+          ])
+
+    Args:
+        name (str): Name of the instance. Overrides parent value.
+        command (str): Command to run in the virtual environment. Overrides parent value.
+        pys  (List[float]): Python versions. Overrides parent value.
+        pkgs (Dict[str, Union[str, List[str]]]): Packages and version(s) to install into the virtual env. Merges and overrides parent values.
+        env  (Dict[str, Union[str, List[str]]]): Environment variables to define in the virtual env. Merges and overrides parent values.
+        venvs (List[Venv]): List of Venvs that inherit the properties of this Venv (unless they are overridden).
```
jd, line 100: It this valid rst/Sphinx format?
(first time I see it :))

Kyle-Verhoog: Yes! It's https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html

docs/quickstart.rst
Kyle-Verhoog: The child overrides the parent value so it would result in just `pys=[3.8]`

docs/conf.py
Kyle-Verhoog: Yes, good catch!

majorgreys (approved):
docs look 👍

PR #88, opened 2020-11-25
Add option for selecting venvs by pattern

### LINKED ISSUE #89: Run venvs based on a matching pattern
One aspect of broader improvements in #24

majorgreys:
Resolves #89

### DIFF
--- a/riot/riot.py
+++ b/riot/riot.py
@@ -170,6 +170,7 @@ class Session:
     def run(
         self,
         pattern: t.Pattern[str],
+        venv_pattern: t.Pattern[str],
         skip_base_install: bool = False,
         recreate_venvs: bool = False,
         out: t.TextIO = sys.stdout,
@@ -199,17 +200,19 @@ class Session:
             }
 
             if pkgs:
-                # Strip special characters for the venv directory name.
-                venv_postfix = "_".join(
-                    [f"{n}{rmchars('<=>.,', v)}" for n, v in pkgs.items()]
-                )
-                venv_name = f"{base_venv}_{venv_postfix}"
+                venv_name = get_venv_directory_name(base_venv, pkgs)
                 pkg_str = " ".join(
                     [f"'{get_pep_dep(lib, version)}'" for lib, version in pkgs.items()]
                 )
             else:
                 venv_name = base_venv
                 pkg_str = ""
+
+            if not venv_pattern.search(venv_name):
+                logger.debug(
+                    "Skipping venv instance '%s' due to pattern mismatch", venv_name
+                )
+                continue
 
             # Result which will be updated with the test outcome.
             result = VenvInstanceResult(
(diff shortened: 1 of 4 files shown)


tests/test_cli.py
```diff
@@ -238,11 +245,71 @@ def test_run_with_pattern(cli: click.testing.CliRunner) -> None:
...
+            result = cli.invoke(
+                riot.cli.main,
+                [
+                    "run",
+                    "test",
+                    "-d",
+                    "--skip-base-install",
+                ],
+            )
+            # Success, but no output because we mock run
```
Kyle-Verhoog, line 267:
```suggestion
```

tests/test_cli.py
```diff
@@ -238,11 +245,71 @@ def test_run_with_pattern(cli: click.testing.CliRunner) -> None:
...
+                [
+                    "run",
+                    "test",
+                    "-d",
+                    "--skip-base-install",
+                    "--venv-pattern",
+                    "pytest543$",
+                ],
+            )
+            # Success, but no output because we mock run
```
Kyle-Verhoog, line 297:
```suggestion
```

tests/test_cli.py
```diff
@@ -238,11 +245,71 @@ def test_run_with_pattern(cli: click.testing.CliRunner) -> None:
...
+                    "-d",
+                    "--skip-base-install",
+                ],
+            )
+            # Success, but no output because we mock run
+            assert result.exit_code == 0
+            assert result.stdout == ""
+
+            mock_debug.assert_called()
+            assert not any(
```
Kyle-Verhoog, line 272: Ideally this should be replaced with checking the `result.stdout` above. Once we move to using click.echo for program printing we should be able to change these over.

riot/riot.py
```diff
@@ -360,6 +368,12 @@ def get_pep_dep(libname: str, version: str) -> str:
     return f"{libname}{version}"
 
 
+def get_venv_directory_name(base_venv, pkgs):
```
Kyle-Verhoog, line 371:
```suggestion
def get_venv_directory_name(base_venv: str, pkgs: t.Dict[str, t.List[str]]) -> str:
```

riot/riot.py
```diff
@@ -293,10 +296,15 @@ def run(
...
             sys.exit(1)
 
-    def list_venvs(self, pattern, pythons=None, out=sys.stdout):
+    def list_venvs(self, pattern, venv_pattern, pythons=None, out=sys.stdout):
         for inst in self.venv.instances(pattern=pattern):
             if pythons and inst.py not in pythons:
                 continue
+            base_venv = get_base_venv_path(inst.py)
+            if not venv_pattern.search(
+                get_venv_directory_name(base_venv, dict(inst.pkgs))
```
Kyle-Verhoog, line 305:
```suggestion
                get_venv_directory_name(base_venv, inst.pkgs)
```

Or was there a reason to make a copy?

riot/riot.py
```diff
@@ -293,10 +296,15 @@ def run(
...
+            ):
+                continue
```
Kyle-Verhoog, line 307: Maybe worth doing a debug log here

riot/riot.py
majorgreys: Problem here is with how this function is called in `Session.run` where `pkgs` is a dict rather than a tuple list. We can just adopt the dict for this in `VenvInstance` as well.

Kyle-Verhoog: Oh I see. Right, because we strip out `None` packages.. I forgot about that functionality.

Kyle-Verhoog: Fine to leave it as-is then IMO

Kyle-Verhoog (approved):

PR #96, opened 2020-11-26
Remove callable env values

Kyle-Verhoog:
This feature breaks the declarative nature of the virtual env building.
If the need for it arises in the future then we can reconsider having
this feature.


jd (approved):
I don't like those `AttrDict` thing 👍

PR #100, opened 2020-11-28
Add test helpers, rewrite integration tests

Kyle-Verhoog:
Add test helpers to do proper input/output testing.


riotfile.py
```diff
@@ -12,7 +12,7 @@
             },
             pys=[3.6, 3.7, 3.8, 3.9],
             pkgs={
-                "pytest": latest,
+                "pytest": "==6.1.2",
```
jd, line 15: tbh not a fan of pinning everything.
Who's gonna upgrade those?

tests/proctest.py
```diff
@@ -0,0 +1,73 @@
...
+import subprocess
+import sys
+import tempfile
+import typing as t
+
+
+if t.TYPE_CHECKING or sys.version_info[:2] >= (3, 9):
+    _T_CompletedProcess = subprocess.CompletedProcess[str]
+else:
+    _T_CompletedProcess = subprocess.CompletedProcess
```
jd, line 12: I think we define this in riot you should be able to import it here

tests/proctest.py
```diff
@@ -0,0 +1,73 @@
...
+        return self.proc.stderr
+
+    @property
+    def returncode(self) -> int:
+        return self.proc.returncode
+
+
+class TestDir:
+    def __init__(self):
+        self._tmpdir = tempfile.TemporaryDirectory()
```
jd, line 36: pytest has a fixture for this
it's not clear who delete the tmpdir here.

tests/proctest.py
```diff
@@ -0,0 +1,73 @@
...
+            f.write(contents)
+
+    def run(
+        self,
+        args: t.Union[str, t.Sequence[str]],
+        env: t.Optional[t.Dict[str, str]] = None,
+        cwd: t.Optional[_T_Path] = None,
+    ) -> ProcResult:
+        if isinstance(args, str):
+            args = args.split(" ")
```
jd, line 63: This should not be needed, right?

tests/proctest.py
```diff
@@ -0,0 +1,73 @@
...
+if t.TYPE_CHECKING or sys.version_info[:2] >= (3, 9):
+    _T_CompletedProcess = subprocess.CompletedProcess[str]
+else:
+    _T_CompletedProcess = subprocess.CompletedProcess
+
+_T_Path = t.Union[str, "os.PathLike[t.Any]"]
+
+
+@dataclasses.dataclass(frozen=True)
+class ProcResult:
```
jd, line 18: not sure why we ned this class?

jd (changes requested):

riotfile.py
Kyle-Verhoog: I think I pinned it because some of the tests assert on the pytest output format which could change in newer versions of pytest

tests/proctest.py
Kyle-Verhoog: I think `subprocess.run` interprets a given string as the executable name. So if you pass `riot run ...` it'll look for an executable called `"riot run ..."` 😕

brettlangdon: Unless you do `subprocess.run(cmd, shell=True)` which is what we do for running venv commands with `source ...`

riotfile.py
Kyle-Verhoog: I'm ok to not pin it - just might make the tests brittle. The better approach might not be dependent on `pytest` in the tests...

tests/test_integration.py
```diff
@@ -0,0 +1,520 @@
...
+
+
+_T_Path = Union[str, "os.PathLike[Any]"]
+
+
+def run(
+    args: Union[str, Sequence[str]], cwd: _T_Path, env: Optional[Dict[str, str]] = None
+) -> _T_CompletedProcess:
+    if isinstance(args, str):
+        args = args.split(" ")
```
brettlangdon, line 22: You can probably do:

```python
subprocess.run(args, shell=isinstance(args, str))
```

tests/test_integration.py
```diff
@@ -0,0 +1,520 @@
...
+        args,
+        env=env,
+        encoding=sys.getdefaultencoding(),
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        cwd=cwd,
+    )
+
+
+def test_no_riotfile(tmp_path: pathlib.Path) -> None:
```
brettlangdon, line 34: where does `tmp_path` come from? existing `pytest` fixture?

tests/test_integration.py
```diff
@@ -0,0 +1,520 @@
...
+Commands:
+  generate  Generate base virtual environments.
+  list      List all virtual env instances matching a pattern.
+  run       Run virtualenv instances with names matching a pattern.
+""".lstrip()
+    )
+    assert result.stderr == ""
+    assert result.returncode == 0
+
+    result = run("riot list", cwd=tmp_path)
```
brettlangdon, line 57: we should get the same for `riot run` and `riot generate` right?

tests/test_integration.py
```diff
@@ -0,0 +1,520 @@
...
+    rf_path = tmp_path / "riotfile.py"
+    rf_path.write_text(
+        """
+from riot import Venv
+venv = Venv(
+    name="test",
+    pys=[3],
+    command="echo hi",
+)
+""",
```
brettlangdon, line 191: I wonder if we should have a bunch of these as pytest fixtures, like:

```python
import pytest

@pytest.fixture
def riotfile_echo_hi(tmp_path: pathlib.Path) -> pathlib.Path:
    rf_path = tmp_path / "riotfile.py"
    rf_path.write_text(""" .... """)
    return rf_path
```


Could even consider the same for something like:

```python
@pytest.fixture
def run_riot(tmp_path: pathlib.Path) -> typing.Callable[[...] _T_CompletedProcess]:
    def wrapper(*args, **kwargs):
        return run(*args, **kwargs, cwd=tmp_path)
    return wrapper


def test_riot(run_riot) -> None:
    result = run_riot("riot list")
```


might be getting a bit carried away with these suggestions, but looking like a lot of repeated patterns in this test file.

tests/test_integration.py
```diff
@@ -0,0 +1,520 @@
...
+    }
+)
+""",
+    )
+    result = run("riot list", cwd=tmp_path)
+    assert result.stderr == ""
+    assert re.search(
+        r"""
+test  .* 'pkg1==1.0'
+test  .* 'pkg1==2.0'
```
brettlangdon, line 216: We should know this exact string no? since this is the python version and we know the currently running version?

tests/test_integration.py
Kyle-Verhoog: Yes

tests/test_integration.py
Kyle-Verhoog: Yeah I was thinking about the second option because of the repeated `cwd=tmp_path` - I think it's worth doing

tests/test_integration.py
Kyle-Verhoog: The reasoning here is that the interpreter could be platform dependent. Right now it'd be fine because the output there would probably just be `Interpreter(_hint="3")` but in #137 we'd also have the path to the interpreter.

What we can do is at least match on `Interpreter(.*)` or similar. WDYT?

tests/test_integration.py
Kyle-Verhoog: https://docs.pytest.org/en/6.2.x/tmpdir.html

tests/test_integration.py
Kyle-Verhoog: Correct, are you suggesting that we should also check those here? Would be an easy parametrization

tests/proctest.py
Kyle-Verhoog: Done in f322ec5

tests/test_integration.py
Kyle-Verhoog: Done in 7b2bb88. I'm personally ok with the boilerplate of creating a riotfile each time. If we need a common base one for a bunch of cases then we can create that fixture when needed.

tests/test_integration.py
```diff
@@ -0,0 +1,540 @@
...
+    def _run(
+        args: Union[str, Sequence[str]],
+        cwd: Optional[_T_Path] = None,
+        env: Optional[Dict[str, str]] = None,
+    ) -> _T_CompletedProcess:
+        if cwd is None:
+            cwd = tmp_path
+        return run(args, cwd, env)
+
+    yield _run
```
brettlangdon, line 58: I don't think this needs to be a generator, you can just do `return _run`

Generator fixtures are good for things that need setup/teardown

(Unless, tmp_path gets torn down on return? I would have assumed fixture teardown happens after each test finishes, not after the scope of where it is used closes?)

tests/test_integration.py
```diff
@@ -0,0 +1,540 @@
...
+    assert (
+        """
+SyntaxError: invalid syntax
+""".lstrip()
+        in result.stderr
+    )
+    assert result.returncode == 1
+
+
+def test_help(tmp_path: pathlib.Path, tmp_run: _T_TmpRun) -> None:
```
brettlangdon, line 152:
```suggestion
def test_help(tmp_run: _T_TmpRun) -> None:
```

tests/test_integration.py
```diff
@@ -0,0 +1,540 @@
...
+  generate  Generate base virtual environments.
+  list      List all virtual env instances matching a pattern.
+  run       Run virtualenv instances with names matching a pattern.
+""".lstrip()
+    )
+    assert result.stderr == ""
+    assert result.returncode == 0
+
+
+def test_version(tmp_path: pathlib.Path, tmp_run: _T_TmpRun) -> None:
```
brettlangdon, line 176:
```suggestion
def test_version(tmp_run: _T_TmpRun) -> None:
```

tests/test_integration.py
```diff
@@ -0,0 +1,540 @@
...
+    rf_path.write_text(
+        """
+from riot import Venv
+venv = Venv(
+    name="test",
+    pys=[3],
+    command="echo hi",
+)
+""",
+    )
```
brettlangdon, line 219: This still feels like a lot of boilerplate, but I am ok with it.

tests/test_integration.py
brettlangdon: Makes me wonder if we need better output here than what we have now `repr(Interpreter)`

could make this more of a consistent format

brettlangdon: not sure I am suggesting a specific change right now, just talking out loud.

tests/test_integration.py
```diff
@@ -0,0 +1,540 @@
...
+    result = tmp_run("riot run test")
+    assert 'ERROR: File "setup.py"' in result.stderr
+    assert "Dev install failed, aborting!" in result.stderr
+    assert result.stdout == ""
+    assert result.returncode == 1
+
+
+def test_bad_interpreter(tmp_path: pathlib.Path, tmp_run: _T_TmpRun) -> None:
```
brettlangdon, line 520:
```suggestion
@with_riotfile(
    """
from riot import Venv
venv = Venv(
    ...
)
    """
    , filename="bad-riotfile.py",  # Obviously the default is `riotfile.py`
)
def test_bad_interpreter(riotfile: pathlib.Path, tmp_run: _T_TmpRun) -> None:
```

```suggestion
@with_riotfile_fixture("some-file-in-data-dir.py)
def test_bad_interpreter(riotfile: pathlib.Path, tmp_run: _T_TmpRun) -> None:
```


Just writing something down as a trial to see how it felt, not suggesting a change here.

tests/test_integration.py
Kyle-Verhoog: The only boilerplate-y thing is the `rf_path = tmp_path / "riotfile.py"` which really isn't that bad IMO. The alternative you have below `@with_riotfile("riotfile.py")` is pretty much the same thing, just declarative rather than imperative.

We could even write these as one liners:

```python
(tmp_path / "riotfile.py").write_text("""
""")
```

tests/test_integration.py
Kyle-Verhoog: 🤷‍♂️ returning equivalent for our purposes but yielding is easier to change if later on we do need to setup/teardown. I don't feel strongly though.

brettlangdon: I don't feel strongly either.

tests/test_integration.py
brettlangdon: yeah, true.

jd (approved):
LGTM.
Adding manual merge in case we want another review. 
Feel free to remove it if everyone's happy.

PR #101, opened 2020-11-30
fix(interpreter): catch the version for Python 2

jd:
Python 2 outputs the version on stderr, not stdout.


majorgreys:
We should add tests to cover this. The existing test for generating base venvs mocks out this code path.

jd:
Agreed though we don't have Python 2 available via GitHub Action I think 😢

majorgreys (approved):

PR #105, opened 2020-12-08
test: add test for env not passed

### LINKED ISSUE #122: Add a test for os.environ passed via pass-env
From @majorgreys in #119
> This looks great! I had forgotten about #105. It would be good here to also have a similar test to what I had added for to confirm the os.environ doesn't get passed along with the instance env you are testing here by default and only does when --pass-env is provided.

majorgreys:
Resolves #122

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -8,6 +8,7 @@ import _pytest.monkeypatch
 import click.testing
 import mock
 import pytest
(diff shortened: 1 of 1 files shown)


jd:
@mergifyio update

jd (approved):

jd:
I guess that needs some fix too :)

brettlangdon (approved):

PR #119, opened 2021-03-17
fix: make sure venv is passed to subprocess.run

jd:
The current code broke the environ passing. Add a test and fix it!

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -496,6 +496,46 @@ venv = Venv(
         assert "✓ success2" in result.stdout
 
 
+def test_env(cli: click.testing.CliRunner) -> None:
+    with cli.isolated_filesystem():
+        with open("riotfile.py", "w") as f:
+            f.write(
+                """
+from riot import Venv, latest
+
+venv = Venv(
+    pkgs={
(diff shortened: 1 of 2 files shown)


jd:
I just see @majorgreys tried https://github.com/DataDog/riot/pull/105 !

P403n1x87 (dismissed):

tests/test_cli.py
```diff
@@ -495,6 +495,46 @@ def test_types(cli: click.testing.CliRunner) -> None:
...
+                """
+import os
+
+def test_success():
+    assert os.environ["foobar"] == "baz"
+            """
+            )
+
+        result = cli.invoke(
+            riot.cli.main, ["run", "-s", "envtest"], catch_exceptions=False
```
majorgreys, line 532: This looks great! I had forgotten about #105. It would be good here to also have a similar test to what I had added for to confirm the `os.environ` doesn't get passed along with the instance env you are testing here by default and only does when `--pass-env` is provided.

jd: #122

Yun-Kim (approved):

PR #121, opened 2021-03-17
ci: do not set LANG

jd:


jd:
Not sure why this is here and Python 3.6 complains, so trying stuff…

Kyle-Verhoog (approved):

Kyle-Verhoog:
Yeah I forget why it was added 🤷‍♂️

riotfile.py
```diff
@@ -6,10 +6,6 @@
         Venv(
             name="test",
             command="pytest {cmdargs}",
-            env={
-                "LC_ALL": "C.UTF-8",
-                "LANG": "C.UTF-8",
```
brettlangdon, line 11: Shouldn't this be `"LANG": "Python"`?

:trollface:

PR #124, opened 2021-03-18
ci: enable isort

jd:


Kyle-Verhoog (approved):

P403n1x87 (approved):

jd:
I don't get why this is failing 😢

P403n1x87:
> I don't get why this is failing 😢

~~~
./riotfile.py:2:1: I100 Import statements are in the wrong order. 'from riot import latest' should be before 'from riot import Venv'
~~~

This would have been valid before, but it seems that isort is not caring about case?

~~~
./tests/test_cli.py:10:1: I202 Additional newline in a group of imports. 'import riot.cli' is identified as Third Party and 'import pytest' is identified as Third Party.
~~~

It looks like isort has its own way of identifying TP imports, so now these should be grouped together.

Introducing isort at this point might imply a significant reordering of dependencies; on the plus side, its lint messages are clearer to address in my experience, so could still be worth going for it in the long run

PR #125, opened 2021-03-18
ci(mergify): simplify merging condition

jd:
No need for a label, just wait for the number of reviews and branch protection
settings.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -2,7 +2,8 @@ pull_request_rules:
   - name: automatic merge
     conditions:
       - base=master
-      - label=automerge
+      - label!=manual merge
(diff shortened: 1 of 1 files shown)


Kyle-Verhoog (approved):

.mergify.yml
```diff
@@ -2,7 +2,8 @@ pull_request_rules:
   - name: automatic merge
     conditions:
       - base=master
-      - label=automerge
+      - label!=manual merge
```
Kyle-Verhoog, line 5: does github actually support whitespace in labels?

jd: of course. and unicode!

PR #128, opened 2021-03-22
fix(riot): replace activate via env var

jd:


jd:
This does not work because it triggers a bug I reported to @Kyle-Verhoog a few weeks ago: packages are not installed in the right venv.

This is a critical issue because it means the venv might just be working by luck in some cases and we might miss cases where dependencies are missing in venv declaration 😢

brettlangdon:
@jd yeah, I think I remember hitting this bug myself not too long ago. It is mostly a side-effect of the fact that we `cp -R` venvs around, some of the links aren't getting updated properly or something.

Do we have a GH Issue tracking that problem already? Do we have a place to discuss it further?

P403n1x87 (approved):

P403n1x87:
@brettlangdon do you reckon this could have been fixed by the latest release 0.8.0, which introduces the `create` argument?

brettlangdon:
> @brettlangdon do you reckon this could have been fixed by the latest release 0.8.0, which introduces the `create` argument?

This might be solved by the prefix work. We can close for now and re-evaluate if the issue crops up again.

PR #132, opened 2021-04-10
Don't copy parent venv, use PYTHONPATH instead

brettlangdon:
We can modify `PYTHONPATH` instead of copying the parent virtualenv.

This might help solve some of the issues we've seen come up with copying them.

I did not put this full a thorough set of tests yet, other than:

1. Base venv can import the source/develop package that we build
2. Installing a dependency only in the child venv is imported properly
3. Installing a dependency only in the parent venv is imported properly from the child


riot/riot.py
```diff
@@ -109,6 +109,16 @@ def venv_path(self) -> str:
         version = self.version().replace(".", "")
         return f".riot/venv_py{version}"
 
+    def site_packages_path(self) -> str:
```
P403n1x87, line 112: can this be tested?

brettlangdon: ok, I tried, I don't think I _love_ the way the test is setup, would love feedback on how to better approach.

jd (approved):

Yun-Kim (approved):

brettlangdon (commented):
The docs test is failing for legitimate reasons. Due to PYTHONPATH issues, so actually great catch.

riot/riot.py
```diff
@@ -86,6 +86,10 @@ def version(self) -> str:
         version = output.decode().strip().split(" ")[1]
         return version
 
+    @functools.lru_cache()
```
jd, line 89: https://docs.python.org/3/library/functools.html#functools.cached_property :(

jd (approved):

tests/test_unit.py
```diff
@@ -26,3 +35,49 @@ def test_interpreter(v1, v2, equal):
...
+
+
+def test_sitepackages_path(current_interpreter: Interpreter) -> None:
+    py_dot_version = ".".join((str(_) for _ in sys.version_info[:2]))
+
+    expected = os.path.join(
+        current_interpreter.venv_path(),
+        "lib/python{}/site-packages".format(py_dot_version),
+    )
+    assert current_interpreter.site_packages_path().endswith(expected)
```
P403n1x87, line 52: Might be worth using `os.path.join` to also concatenate the bits in `lib/python{}/site-packages`, and perhaps convert to an absolute path with `os.path.abs` and do exact matching, in case this will ever need to be run on other platforms, e.g. win?

brettlangdon: yeah, join works well, good suggestion

riot/riot.py
```diff
@@ -86,6 +86,15 @@ def version(self) -> str:
...
+    def version_info(self) -> t.Tuple[int, int, int]:
+        version_head = "".join(
+            list(itertools.takewhile(lambda _: _.isdigit() or _ == ".", self.version()))
+        )
+        # Return (major, minor, patch)
+        version_info = tuple(int(_) for _ in version_head.split(".")) + (0, 0, 0)
+        return (version_info[0], version_info[1], version_info[2])
```
Kyle-Verhoog, line 96:
```suggestion
        return version_info
```
?

riot/riot.py
```diff
@@ -109,6 +118,17 @@ def venv_path(self) -> str:
         version = self.version().replace(".", "")
         return f".riot/venv_py{version}"
 
+    def site_packages_path(self) -> str:
```
Kyle-Verhoog, line 121: This function feels a little brittle. Is there a way to get the site packages path from the interpreter itself?

A quick google suggests that maybe `python -c 'import site; print(site.getsitepackages())'` would do the trick

riot/riot.py
```diff
@@ -241,6 +261,17 @@ def venv_path(self) -> str:
         venv_postfix = "_".join([f"{n}{rmchars('<=>.,', v)}" for n, v in self.pkgs])
         return f"{base_path}_{venv_postfix}"
 
+    def site_packages_path(self) -> str:
```
Kyle-Verhoog, line 264: This is the same method as defined above? Surprised linting doesn't catch this lol

riot/riot.py
```diff
@@ -374,16 +409,35 @@ def run(
...
+            if pass_env:
+                env = os.environ.copy()
+            else:
+                env = {}
+
+            # Add in the instance env vars.
+            env.update(dict(inst.env))
+
+            if python_path:
+                if "PYTHONPATH" in os.environ:
```
Kyle-Verhoog, line 422: Do we actually want this? I feel like it could lead to some baffling behaviour as a user because packages I don't specify in my riot virtualenv might be locatable through a sneaky PYTHONPATH environment variable being set in my shell.

Did you have a use-case in mind here? I think there might be cases for it but I'd prefer it be opt-in.

riot/riot.py
```diff
@@ -374,16 +409,35 @@ def run(
...
+                    py_ex = inst.py.path()
                     logger.info(
-                        "Copying base virtualenv '%s' into virtualenv '%s'.",
-                        base_venv_path,
-                        venv_path,
+                        "Creating virtualenv '%s' with interpreter '%s'.", py_ex, py_ex
                     )
+
                     try:
-                        shutil.copytree(base_venv_path, venv_path, symlinks=True)
```
Kyle-Verhoog, line 386: 👌

Kyle-Verhoog (changes requested):

riot/riot.py
brettlangdon: I had issues with mypy typing, since I explicitly want a `Tuple[int, int, int]`. If someone has another idea to make it work, I am game.

The type from L95 is `Tuple[int, ...]`

riot/riot.py
brettlangdon: Cool, I do actually like that.

Do you think we should grab all of the site packages directories? Or should it always be 1 element and we can get away with `site.getsitepackages()[0]`?

riot/riot.py
brettlangdon: It is on a different class?

riot/riot.py
brettlangdon: I was only thinking for compatibility, but makes sense to fully isolate and ignore any globally set env variables.

brettlangdon (commented):
At this point I will break this PR up, some changes to the version parsing/caching, and then the site packages stuff from here.

riot/riot.py
brettlangdon: I didn't notice before, this is 3.8 only :(

brettlangdon:
I took the version stuff out in #138.

riot/riot.py
Kyle-Verhoog: I suppose the function could return all since it might be useful in the future.

riot/riot.py
Kyle-Verhoog: oh... I missed that but my concern was the duplicated code (the bodies of the two methods are essentially the same)

brettlangdon: _almost_ yeah, I'll see if I can refactor a bit with the other suggested change

riot/riot.py
```diff
@@ -381,47 +409,57 @@ def run(
...
                     except FileNotFoundError:
                         logger.info("Base virtualenv '%s' does not exist", venv_path)
                         continue
                     except FileExistsError:
                         # Assume the venv already exists and works fine
                         logger.info("Virtualenv '%s' already exists", venv_path)
 
-                    logger.info("Installing venv dependencies %s.", pkg_str)
+                # DEV: We need the venv to exist first since we query the venv
+                # python for it's site-packages directories
```
jd, line 446:
```suggestion
                # python for its site-packages directories
```

jd (approved):

jd (approved):

Kyle-Verhoog (approved):

PR #135, opened 2021-04-29
use default shell on system

rahul-kumi:
Update the initialization to accept the default shell set in the system, if that's not found to fall back to `/bin/bash`

### DIFF
--- a/releasenotes/notes/use-system-shell-222f6711f7b17dd4.yaml
+++ b/releasenotes/notes/use-system-shell-222f6711f7b17dd4.yaml
@@ -1,0 +1,5 @@
+---
+features:
+  - |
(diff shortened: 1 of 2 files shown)


jd (approved):

brettlangdon (approved):

brettlangdon:
@rahul-kumi thanks for the contribution, the only failing check is for adding a changelog entry.

We use [reno](https://docs.openstack.org/reno/latest/) to manage release notes.

Once you have it installed you can run `reno new <slug>`

Or we have a command in our `riotfile.py` for `riot` for reno, you can run `riot run -s reno new <slug>`.

brettlangdon (approved):

PR #136, opened 2021-04-29
Add support for CIRCLE_NODE_{TOTAL,INDEX} environment variables

### LINKED ISSUE #47: Support CircleCI env variables for parallelism
https://circleci.com/docs/2.0/parallelism-faster-jobs/#using-environment-variables-to-split-tests

If we can detect and use these to auto-segment envs, then our parallelism support in CircleCI will be more magical.

You will only need to set the `parallelism: #` setting.

brettlangdon:
Fixes #47

I am not convinced having this be the default behavior is the best? Should we require a cli switch to say to use these env variables, or maybe an env variable to disable this behavior?

can we think of a case when someone would not want this? only thing I can think of is if you are using some other tooling to split tests envs that you give to riot, like https://github.com/DataDog/dd-trace-py/blob/8ded786edb6e0b3d7b2e079ae58a0d9ab87ffd05/.circleci/config.yml#L130

### DIFF
--- a/riot/riot.py
+++ b/riot/riot.py
@@ -4,6 +4,7 @@ import importlib.abc
 import importlib.util
 import itertools
 import logging
+import math
 import os
 import shutil
 import subprocess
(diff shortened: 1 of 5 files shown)


riot/riot.py
```diff
@@ -309,6 +312,55 @@ def is_warning(self, output):
...
+
+        ci_node_total = os.environ.get("CI_NODE_TOTAL")
+        ci_node_index = os.environ.get("CI_NODE_INDEX")
+        if os.environ.get("CIRCLECI") == "true":
+            if "CIRCLE_NODE_TOTAL" in os.environ and "CIRCLE_NODE_INDEX" in os.environ:
+                ci_node_total = os.environ["CIRCLE_NODE_TOTAL"]
+                ci_node_index = os.environ["CIRCLE_NODE_INDEX"]
+
+        if ci_node_total and ci_node_index:
+            venvs = list(_iter_venvs())
```
brettlangdon, line 355: I was originally thinking we might want to do `venvs = sorted(_iter_venvs())` to ensure the results are consistent, but would it be weird that the ordering of the venvs changes based on whether you use parallelism or not?

PR #137, opened 2021-05-15
Refactor interpreter evaluation

Kyle-Verhoog:
Before interpreter evaluation was done at run-time when the interpreter
was first used.

With this patch interpreters are looked up when the environments are
generated.

Also more precise interpreter specification is possible. Before,
specifying "3" in a riotfile would result in a "python3" executable
being looked up. So riot would fail if Python was installed as "python"
and no "python3" executable was available on the path.

Now the path is searched for python executables which are then
referenced when resolving the python to use for an environment.


riot/riot.py
```diff
@@ -62,60 +66,157 @@ def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
...
+        )
+        version = Version(output.decode().strip().split(" ")[1])
+        self._version = version
 
-    def __post_init__(self, hint: _T_hint) -> None:
-        """Normalize the data."""
-        self._hint = str(hint)
+    def __hash__(self) -> int:
+        """Return the hash of this interpreter."""
+        return hash(self._path)
```
jd, line 84: Don't do this, just pass `hash=False` to `_version` L72.

riot/riot.py
```diff
@@ -62,60 +66,157 @@ def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
...
+
+    def __repr__(self) -> str:
+        """Return the repr containing the path and version of this interpreter."""
+        return f"{self.__class__.__name__}('{self.version}', '{self._path}')"
 
     def __str__(self) -> str:
-        """Return the path of the interpreter executable."""
+        """Return the repr of this interpreter."""
         return repr(self)
```
jd, line 92: ditto, not worth it I guess with dataclass doing the job.

riot/riot.py
```diff
@@ -62,60 +66,157 @@ def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
...
 
+    @staticmethod
+    def _matches(v1: Version, v2: Version) -> bool:
+        """Return if v2 matches v1.
```
jd, line 96: It took me a few seconds to understand why that was here.

You should be a little more explicit about what you want to achieve with this one. I understand it makes sense for our use case, but it was not obvious at first read. 😉 

A "cleaner" alternative might be to have a custom `Version` class that inherits from the `packaging` `Version` and that has a new `matches` method like this one. If that's doable, it'd prevent to have this class "polluted" with this matching code altogether.

riot/riot.py
```diff
@@ -62,60 +66,157 @@ def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
...
+        False
+        >>> Interpreter._matches(Version("3.8.2"), Version("2"))
+        False
+        >>> Interpreter._matches(Version("3.8.2-dev"), Version("3.8.2"))
+        True
+        >>> Interpreter._matches(Version("3.8.2-alpha"), Version("3.8.2"))
+        True
+
+        FIXME: unfortunately Version will set undefined segments to 0 so
+        we cannot distinguish between "3.0.0" and "3".
```
jd, line 118: I don't think it's a problem, in general, and especially in our case.

riot/riot.py
```diff
@@ -62,60 +66,157 @@ def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
...
-    def version(self) -> str:
-        path = self.path()
+    def find(cls) -> t.List["Interpreter"]:
+        """Find all Python interpreters discoverable on the PATH."""
+        ex_dirs = os.environ.get("PATH", "").split(":")
+        # Use a list because order matters (need the first match).
+        interp_paths: t.List[str] = []
+        # packaging says that the regular expression needs to be compiled
+        # with the following flags
+        expr = re.compile(f"python{VERSION_PATTERN}$", re.IGNORECASE | re.VERBOSE)
```
jd, line 143: define this has a class variable to compile once and for all

riot/riot.py
```diff
@@ -62,60 +66,157 @@ def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
...
+        for d in ex_dirs:
+            if not os.path.isdir(d):
+                continue
+            for f in os.listdir(d):
```
jd, line 147: As a rule of thumb, never do that. The pattern of calling `stat` before doing an I/O is at best unoptimized, at worst a security issue.

You should always do a single file operation and catch errors, like this:

```suggestion
        try:
            for f in os.listdir(d):
             [...]
        except NotADirectoryError:
             continue
```

riot/riot.py
```diff
@@ -62,60 +66,157 @@ def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
...
+    @classmethod
+    def find_match(cls, v: Version) -> "Interpreter":
+        """Return a matching interpreter for the given version.
+
+        If one is not found then raises a FileNotFoundError.
+        """
+        interps = [i for i in cls.find() if i.matches(v)]
+        if len(interps):
+            return interps[0]
+        raise FileNotFoundError("No interpreter matching %r", v)
```
jd, line 173: There's no need to build the entire list. If it huges, it'll take hours for nothing.
You could use `first` which I like for this: https://pypi.org/project/first/
Otherwise:

```suggestion
        for interp in cls.find():
            if interp.matches(v):
                return interp
        raise FileNotFoundError("No interpreter matching %r", v)
```

riot/riot.py
```diff
@@ -62,60 +66,157 @@ def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
...
+        ex = shutil.which(s)
+        if ex:
+            return cls(ex)
 
-        if py_ex:
-            return py_ex
+        try:
+            version = Version(s)
+        except InvalidVersion as e:
+            raise FileNotFoundError("No interpreter found for %r" % s) from e
```
jd, line 192: I wouldn't change that exception, it does not seem fair.

riot/riot.py
```diff
@@ -62,60 +66,155 @@ def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
...
 class Interpreter:
-    _T_hint = t.Union[float, int, str]
+    _path: str = dataclasses.field()
+    _version: Version = dataclasses.field(init=False)
+    _version_expr = re.compile(f"python{VERSION_PATTERN}$", re.IGNORECASE | re.VERBOSE)
+
+    def __post_init__(self):
+        """Get the version of the interpreter."""
+        output = subprocess.check_output(
+            [self._path, "--version"], stderr=subprocess.STDOUT
```
brettlangdon, line 78: you might want to rebase on `master` we made some changes to how we get version from Python to alleviate the parsing of `--version` output.

https://github.com/DataDog/riot/blob/6d064caf94a51a4ce41600b1e0090004510a9567/riot/riot.py#L81-L91

riot/riot.py
brettlangdon: The better approach is to have a custom `Version` class that inherits from `packaging.version.Version` and which defines a custom `Version._key`.

https://github.com/pypa/packaging/blob/47d40f640fddb7c97b01315419b6a1421d2dedbb/packaging/version.py#L292-L299

Seems like we want to compare versions but ignoring pre/post/dev/local ?

which if that is the case, then a custom key, and then relying on the built-in comparisons to `Version` should work.

riot/riot.py
```diff
@@ -62,60 +66,155 @@ def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
...
+    @property
+    def version(self) -> Version:
+        return self._version
 
-        raise FileNotFoundError(f"Python interpreter {self._hint} not found")
+    def matches(self, version: Version) -> bool:
+        """Return whether the given version matches the version of the interpreter."""
+        return self._matches(self.version, version)
+
+    def path(self) -> str:
```
brettlangdon, line 202:
```suggestion
    @property
    def path(self) -> str:
```

?

riot/riot.py
```diff
@@ -233,20 +346,26 @@ class VenvInstance:
     env: t.Tuple[t.Tuple[str, str]]
     name: t.Optional[str]
     pkgs: t.Tuple[t.Tuple[str, str]]
-    py: Interpreter
+    interpreter_hint: str
+    """A null interpreter means that it was not found."""
```
brettlangdon, line 350:
```suggestion
    # A null interpreter means that it was not found.
```

?

riot/riot.py
```diff
@@ -332,49 +461,55 @@ def run(
...
+                base_venv_path = inst.interpreter.venv_path()
 
-            # Resolve the packages required for this instance.
-            pkgs: t.Dict[str, str] = {
-                name: version for name, version in inst.pkgs if version is not None
-            }
+                # Resolve the packages required for this instance.
+                pkgs: t.Dict[str, str] = {
+                    name: version for name, version in inst.pkgs if version is not None
+                }
```
brettlangdon, line 489: I found transforming/relocating this and `pkg_str` into `VenvInstance` much nicer to work with.

riot/riot.py
```diff
@@ -479,19 +611,32 @@ def run(
...
+        pythons: t.Optional[t.Set[Interpreter]] = None,
+    ) -> None:
         for inst in self.venv.instances(pattern=pattern):
-            if pythons and inst.py not in pythons:
+            if pythons and inst.interpreter not in pythons:
+                continue
+            if not inst.interpreter:
+                pkgs_str = " ".join(
+                    f"'{get_pep_dep(name, version)}'" for name, version in inst.pkgs
+                )
```
brettlangdon, line 626: relocating this to `VenvInstance` will definitely be nice since it seems like we do this transformation in at least 3 places.

riot/riot.py
```diff
@@ -479,19 +611,32 @@ def run(
...
+                env_str = env_to_str(inst.env)
+                click.echo(
+                    f"{inst.name} {env_str} <Interpreter missing for '{inst.interpreter_hint}'> {pkgs_str}"
```
brettlangdon, line 629: should the str/repr output of `Interpreter` just state whether it exists or not?

riot/riot.py
```diff
@@ -479,19 +611,32 @@ def run(
...
             if not venv_pattern.search(inst.venv_path()):
                 continue
+
             pkgs_str = " ".join(
                 f"'{get_pep_dep(name, version)}'" for name, version in inst.pkgs
             )
             env_str = env_to_str(inst.env)
-            py_str = f"Python {inst.py}"
-            click.echo(f"{inst.name} {env_str} {py_str} {pkgs_str}")
+            click.echo(f"{inst.name} {env_str} {inst.interpreter} {pkgs_str}")
```
brettlangdon, line 639: should we make `VenvInstance.__str__` print what we expect here?

or else a helper method would help since we have this format in at least 2 places

riot/riot.py
```diff
@@ -502,25 +647,28 @@ def generate_base_venvs(
...
-            [inst.py for inst in self.venv.instances(pattern=pattern)]
-        )
-        # Apply Python filters.
-        if pythons:
-            required_pys = required_pys.intersection(pythons)
+        required_pys = []
+
+        for inst in self.venv.instances(pattern=pattern):
+            if inst.interpreter is None:
+                continue
```
brettlangdon, line 654: raise?

wondering how we can end up in this state other than misconfiguration?

PR #138, opened 2021-05-20
Add Interprter.version_info and version test cases

brettlangdon:
Add new `Interpreter.version_info` function and some version test cases.

### DIFF
--- a/riot/riot.py
+++ b/riot/riot.py
@@ -81,19 +81,21 @@ class Interpreter:
     def version(self) -> str:
         path = self.path()
 
-        # Redirect stderr to stdout because Python 2 prints version on stderr
-        output = subprocess.check_output([path, "--version"], stderr=subprocess.STDOUT)
-        version = output.decode().strip().split(" ")[1]
-        return version
+        output = subprocess.check_output(
+            [
+                path,
+                "-c",
+                'import sys; print("%s.%s.%s" % (sys.version_info.major, sys.version_info.minor, sys.version_info.micro))',
+            ],
+        )
+        return output.decode().strip()
 
     @functools.lru_cache()
     def version_info(self) -> t.Tuple[int, int, int]:
-        version_head = "".join(
-            list(itertools.takewhile(lambda _: _.isdigit() or _ == ".", self.version()))
-        )
-        # Return (major, minor, patch)
-        version_info = (tuple(int(_) for _ in version_head.split(".")) + (0, 0, 0))[:3]
-        return t.cast(t.Tuple[int, int, int], version_info)
+        path = self.path()
+
+        version_info = tuple(int(_) for _ in self.version().split(".")) + (0, 0, 0)
+        return t.cast(t.Tuple[int, int, int], version_info[:3])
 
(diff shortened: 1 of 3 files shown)


riot/riot.py
```diff
@@ -86,6 +86,15 @@ def version(self) -> str:
         version = output.decode().strip().split(" ")[1]
         return version
 
+    @functools.lru_cache()
+    def version_info(self) -> t.Tuple[int, int, int]:
+        version_head = "".join(
+            list(itertools.takewhile(lambda _: _.isdigit() or _ == ".", self.version()))
```
jd, line 92: Can we use something like `python -c 'import sys; print("%s.%s.%s" % (sys.version_info.major, sys.version_info.minor, sys.version_info.micro))'` in  the`version()` and something like that here?

Sounds safer than parsing strings.

brettlangdon: sounds good! on it!

tests/test_unit.py
```diff
@@ -26,3 +34,16 @@ def test_interpreter(v1, v2, equal):
         assert Interpreter(v1) != Interpreter(v2)
         assert hash(Interpreter(v1)) != hash(Interpreter(v2))
         assert repr(Interpreter(v1)) != repr(Interpreter(v2))
+
+
+def test_interpreter_version(current_interpreter: Interpreter) -> None:
+    version, _, _ = sys.version.partition(" ")
```
brettlangdon, line 40:
```suggestion
    version = "%s.%s.%s" % (sys.version_info.major, sys.version_info.minor, sys.version_info.micro)
```

riot/riot.py
```diff
@@ -81,10 +81,19 @@ def __str__(self) -> str:
...
+                "-c",
+                'import sys; print("%s.%s.%s" % (sys.version_info.major, sys.version_info.minor, sys.version_info.micro))',
+            ],
+        )
+        return output.decode().strip()
+
+    @functools.lru_cache()
+    def version_info(self) -> t.Tuple[int, int, int]:
+        version_info = tuple(int(_) for _ in self.version().split(".")) + (0, 0, 0)
+        return t.cast(t.Tuple[int, int, int], version_info[:3])
```
jd, line 96:
```suggestion
        return tuple(map(int, self.version().split("."))
```
no?

brettlangdon: will still need the cast, but yeah, this is cleaner. thanks!

tests/test_unit.py
```diff
@@ -26,3 +34,20 @@ def test_interpreter(v1, v2, equal):
...
+    )
+    assert current_interpreter.version() == version
+
+
+def test_interpreter_version_info(current_interpreter: Interpreter) -> None:
+    assert current_interpreter.version_info() == (
+        sys.version_info[0],
+        sys.version_info[1],
+        sys.version_info[2],
+    )
```
jd, line 53:
```suggestion
    assert current_interpreter.version_info() == sys.version_info[:3]
```

jd (approved):

PR #139, opened 2021-05-20
Add Interpreter.pythonpath and VenvInstance.pythonpath properties

brettlangdon:
This breaks out the new `Interprter.pythonpath` and `VenvInstance.pythonpath` properties from #132

### DIFF
--- a/releasenotes/notes/add-pythonpath-d776c0551e1f22e5.yaml
+++ b/releasenotes/notes/add-pythonpath-d776c0551e1f22e5.yaml
@@ -1,8 +1,6 @@
 ---
 features:
   - |
-    Add ``Interpreter.sitepackages_paths()`` method to get the expected ``site-package`` directories.
-  - |
     Add ``Interpreter.pythonpath`` property to get expected ``PYTHONPATH`` env value for that ``Interpreter``.
   - |
     Add ``VenvInstance.pythonpath`` property to get expected ``PYTHONPATH`` env value for that ``VenvInstance``.
--- a/tests/test_unit.py
+++ b/tests/test_unit.py
@@ -72,10 +72,6 @@ def test_get_venv_sitepackages(current_venv: Venv) -> None:
     assert get_venv_sitepackages(instances[0].venv_path()) == site.getsitepackages()
 
 
-def test_interpreter_sitepackages(current_interpreter: Interpreter) -> None:
-    assert current_interpreter.sitepackages_paths() == site.getsitepackages()
-
-
 def test_interpreter_pythonpath(current_interpreter: Interpreter) -> None:
     assert current_interpreter.pythonpath == ":".join(site.getsitepackages())
 
--- a/tests/test_unit.py
+++ b/tests/test_unit.py
@@ -1,13 +1,25 @@
+import os
+import re
+import site
 import sys
 
+import mock
 import pytest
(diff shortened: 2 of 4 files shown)


tests/test_unit.py
```diff
@@ -46,8 +58,56 @@ def test_interpreter_version(current_interpreter: Interpreter) -> None:
...
-        sys.version_info[2],
-    )
+    assert current_interpreter.version_info() == sys.version_info[:3]
+
+
+def test_get_python_sitepackages() -> None:
+    assert get_python_sitepackages("python") == site.getsitepackages()
+
+
+@pytest.mark.skip(reason="We are unable to execute venv commands in tests")
```
Kyle-Verhoog, line 68: Why is this the case and is there value in keeping this test case if it's not runnable?

tests/test_unit.py
```diff
@@ -46,8 +58,56 @@ def test_interpreter_version(current_interpreter: Interpreter) -> None:
...
+      When no VenvInstance.pkgs are defined
+        Will be the Interpreter.pythonpath
+    """
+    instances = list(current_venv.instances(default_venv_pattern))
+    assert instances
+    for instance in instances:
+        assert instance.pythonpath == instance.py.pythonpath
+
+
+def test_venv_instance_with_pkgs_pythonpath(current_venv: Venv) -> None:
```
Kyle-Verhoog, line 92: I think I'd much rather have integration tests for this sort of thing with real usage (`riot run ...`). While it's great we can test at a high granularity, these kinds of tests are hard to maintain as they're brittle to many sources that we don't intend to test here (eg. we change the `get_venv_sitepackages` name or implementation).

I'd actually be more comfortable with a subprocess test that has riot construct a Venv and then checks to see which virtualenv the package got installed to.

WDYT?

Kyle-Verhoog (commented):
Overall looks good just a couple thoughts/concerns about the testing strategy

tests/test_unit.py
jd: This PR might help https://github.com/DataDog/riot/pull/128 but it needs #132 😢

brettlangdon: I wanted to basically leave it as a note as to why the test case is missing.

I suppose this isn't 100% the case, we'd have to setup the venv in order to test. Which is probably fine as you note in another comment about integration tests. I wasn't thinking of adding that at this time, but no reason not to.

tests/test_unit.py
brettlangdon: yeah, this makes sense. I wasn't thinking about integration tests right now, but probably good thing to figure out now since our test coverage in general is lacking, good to get ahead of it.

releasenotes/notes/add-pythonpath-d776c0551e1f22e5.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    Add ``Interpreter.pythonpath`` property to get expected ``PYTHONPATH`` env value for that ``Interpreter``.
```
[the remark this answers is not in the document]
brettlangdon: That is a good question.... Maybe?

brettlangdon:
I am going to close this and merge the changes into #132, since we are moving towards integration tests for this, it'll be easier to test with the PYTHONPATH changes in #132

PR #141, opened 2021-07-13
refactor: crazy breaking refactor

P403n1x87:
This refactor introduces the concept of Tasks. Key change is the parenting of venv instances which allows determining the `PYTHONPATH` to use when running commands inside nested venvs.

## Potential future improvements

- Add subtasks to tasks
- Add dependencies between tasks

### DIFF
--- a/riot/__init__.py
+++ b/riot/__init__.py
@@ -1,5 +1,7 @@
-__all__ = ["latest", "Session", "Venv"]
+__all__ = ["latest", "Session", "Task", "Venv"]
 
-from .riot import Session, Venv
+from riot.riot import Session
+from riot.task import Task
+from riot.venv import VenvSpec as Venv
 
 latest = ""
--- a/riot/venv.py
+++ b/riot/venv.py
@@ -1,0 +1,222 @@
+import dataclasses
+import logging
+import os.path
+import subprocess
+import typing as t
+
+from riot.interpreter import Interpreter
+from riot.utils import (
+    CmdFailure,
+    expand_specs,
+    get_pep_dep,
+    rm_singletons,
+    rmchars,
+    run_cmd,
+    run_cmd_venv,
+    to_list,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class VenvError(Exception):
+    pass
+
+
+@dataclasses.dataclass
+class Venv:
+    pkgs: t.Tuple[t.Tuple[str, str]]
+    py: t.Optional[Interpreter]
+    parent: t.Optional["Venv"] = None
+
+    @property
+    def venv_path(self) -> t.Optional[str]:
+        """Return path to directory of the venv this instance should use.
+
+        This will return a python version + package specific venv path name.
+
+        If no packages are defined it will return the ``Interpreter.venv_path``.
+        """
+        if self.py is None:
+            return None
+
+        venv_path = self.py.venv_path
+        if self.needs_venv:
+            venv_path = "_".join(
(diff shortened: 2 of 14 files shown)


riot/venv.py
```diff
@@ -0,0 +1,206 @@
...
+        return " ".join(
+            [
+                f"'{get_pep_dep(lib, version)}'"
+                for lib, version in self.pkgs
+                if version is not None
+            ]
+        )
+
+    @property
+    def pythonpath(self):
```
P403n1x87, line 73: Don't mind the debug noise. By walking up the parenting relation we can determine the PYTHONPATH to use in order to import the correct dependencies 🤞

riot/venv.py
```diff
@@ -0,0 +1,206 @@
...
+    ) -> t.Generator[Venv, None, None]:
+        pys = pys or self.pys or [None]
+        for py in pys:
+            for ps in expand_specs(self.pkgs):
+                venv = Venv(
+                    py=py,
+                    pkgs=ps,
+                    parent=parent_venv,
+                )
+                if not self.venvs:
```
P403n1x87, line 198: I think the `if` and the `for` loops should be this way (currently it's inside-out). This way one can set the parenting relation between venv instances.

riot/__init__.py
```diff
@@ -1,5 +1,7 @@
-__all__ = ["latest", "Session", "Venv"]
+__all__ = ["latest", "Session", "Task", "Venv"]
 
-from .riot import Session, Venv
+from riot.riot import Session
+from riot.task import Task
+from riot.venv import VenvSpec as Venv
```
P403n1x87, line 5: If the argument `venvs` of `Task` could be made of type `List[Dict[..., ...]]` then `Venv` wouldn't be required to be part of the public API.

Kyle-Verhoog (commented):
Interestingly enough this brings us roughly back to the early days of riot where it was `Suite` and `Venv` where `Suite` defined the name and command IIRC. Now it's called `Task` which I do like a bit better.

I really like the idea of having a defined top level structure which makes the riotfile more intuitive than all the nested Venv things. Also helps clean up the implementation because we don't need to worry about so many things down the tree.

We could do some of the file break out changes ahead of time which I think would help clean this PR up some.

I'm personally not a fan of having a `utils` file as it just becomes a dumping ground and doesn't really save much clutter. Not a super strong opinion though.


Overall good stuff and I think this is a good improvement!

riot/venv.py
```diff
@@ -0,0 +1,222 @@
...
+
+class VenvError(Exception):
+    pass
+
+
+@dataclasses.dataclass
+class Venv:
+    pkgs: t.Tuple[t.Tuple[str, str]]
+    py: t.Optional[Interpreter]
+    parent: t.Optional["Venv"] = None
```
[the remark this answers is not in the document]
P403n1x87: But tasks can share a venv :)

~~~ python
checks_venv = Venv(
    pkgs = { "black": latest, "isort": latest }
)

tasks = [
    Task(name="fmt", command="black . && isort .", venvs=[checks_venv])
    Task(name="black", command="black .", venvs=[checks_venv])
    Task(name="isort", command="isort .", venvs=[checks_venv])
]
~~~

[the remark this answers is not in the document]
Kyle-Verhoog: oh sure, although it does become more procedural and less declarative/json/yaml-able

P403n1x87:
The main branch has diverged considerably. We will maybe take the general idea and implement it with a new PR in the future, but we might split the `riot.py` source first.

PR #142, opened 2021-07-15
fix: control and track pip install target dirs

P403n1x87:
The logic that infers all the venv instances has been reworked to track parenting of venv instances. Installation of packages via pip now happens in a predefined folder. This helps with deriving the elements that need to go to the `PYTHONPATH` and the `PATH` environment to ensure correct layering of re-used intermediate venvs.

### DIFF
--- a/riot/riot.py
+++ b/riot/riot.py
@@ -63,6 +63,10 @@ def to_list(x: t.Union[_K, t.List[_K]]) -> t.List[_K]:
 _T_stdio = t.Union[None, int, t.IO[t.Any]]
 
 
+class VenvError(Exception):
+    pass
+
+
 @dataclasses.dataclass(unsafe_hash=True, eq=True)
 class Interpreter:
     _T_hint = t.Union[float, int, str]
@@ -97,6 +101,15 @@ class Interpreter:
             t.Tuple[int, int, int], tuple(map(int, self.version().split(".")))
         )
 
+    @property
+    def bin_path(self) -> t.Optional[str]:
+        return os.path.join(self.venv_path, "bin")
+
+    @property
+    def site_packages_path(self) -> t.Optional[str]:
+        version = ".".join((str(_) for _ in self.version_info()[:2]))
+        return os.path.join(self.venv_path, "lib", f"python{version}", "site-packages")
+
     @functools.lru_cache()
     def path(self) -> str:
         """Return the Python interpreter path or raise.
@@ -111,22 +124,19 @@ class Interpreter:
             py_ex = shutil.which(f"python{self._hint}")
 
         if py_ex:
-            return py_ex
+            return os.path.abspath(py_ex)
 
         raise FileNotFoundError(f"Python interpreter {self._hint} not found")
 
+    @property
     def venv_path(self) -> str:
         """Return the path to the virtual environment for this interpreter."""
         version = self.version().replace(".", "")
-        return f".riot/venv_py{version}"
-
-    @property
-    def pythonpath(self) -> str:
-        return ":".join(get_venv_sitepackages(self.venv_path()))
+        return os.path.abspath(f".riot/venv_py{version}")
 
     def create_venv(self, recreate: bool) -> str:
         """Attempt to create a virtual environment for this intepreter."""
-        path: str = self.venv_path()
+        path: str = self.venv_path
 
         if os.path.isdir(path) and not recreate:
             logger.info(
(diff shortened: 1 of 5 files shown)


tests/test_integration.py
```diff
@@ -454,7 +453,10 @@ def test_failure():
     assert result.returncode == 0
 
     result = tmp_run("riot run -s fail")
-    assert "x fail:  pythonInterpreter(_hint='3') 'pytest'\n" in result.stdout
+    print(result.stdout)
```
brettlangdon, line 456:
```suggestion
```

?

brettlangdon (commented):
nice, I like this change

riot/riot.py
```diff
@@ -272,26 +267,101 @@ def needs_venv(self) -> bool:
...
+            if self.needs_venv:
+                py_ex = py.path()
+                logger.info(
+                    "Creating virtualenv '%s' with interpreter '%s'.",
+                    self.venv_path,
+                    py_ex,
+                )
+
+                try:
+                    run_cmd(
```
P403n1x87, line 335: I think there is no need to create more venvs other than the base ones with this change, but I think it's best to experiment with this in a follow-up PR.

riot/riot.py
```diff
@@ -192,78 +204,70 @@ def __post_init__(self, pys, pkgs, env):
...
             venv_path = "_".join(
                 [venv_path] + [f"{n}{rmchars('<=>.,', v)}" for n, v in self.pkgs]
             )
         return venv_path
 
+    @property
+    def target(self) -> t.Optional[str]:
+        if self.venv_path is None:
+            return None
+        return self.venv_path
```
P403n1x87, line 269: 🤦 
```suggestion
        return self.venv_path
```

riotfile.py
```diff
@@ -6,13 +6,22 @@
...
+                "click": latest,
                 "pytest": latest,
                 "pytest-cov": latest,
                 "mock": latest,
                 "typing-extensions": latest,
             },
+            venvs=[
+                Venv(
+                    pys=[3.6],
+                    pkgs={"dataclasses": latest},
```
brettlangdon, line 19: why didn't this fail before?

brettlangdon (dismissed):

riotfile.py
P403n1x87: Was wondering the same thing as the dependency on `click` was also missing. The only explanation that I could think of is that perhaps the site where riot was being installed leaked into the task runs. Now this doesn't seem to happen 🙂

P403n1x87: 🤔  in fact these shouldn't be needed now that we also set the base venv in the `PYTHONPATH`, since those come with the installation of the dev package. Will investigate and fix

P403n1x87: @brettlangdon this is now fixed!

riot/riot.py
brettlangdon: well, that is what we handle with `if self.needs_venv` right? we only need to create additional venvs whenever an intermediary venv has dependencies?

releasenotes/notes/fix-pythonpath-3e26d996b9df139a.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    Fixed layering of package installations by tracking the parent-child
```
Kyle-Verhoog, line 4: Since the change to using PYTHONPATH hasn't actually been released yet I don't think we need the release note. Although this snippet would be great in the commit message as it succinctly describes the fix here :+1:

riot/riot.py
P403n1x87: I think it is enough to run `pip install` within the base environment because we then specify the location of each dependency based on the "venv" configuration with `--prefix`. We can then decide which packages to actually use in the base venv by adding paths to `PYTHONPATH`. So by running inside the base venv with the right `PYTHONPATH` we should be OK.

Kyle-Verhoog (commented):
No release note needed - otherwise LGTM!! great stuff... I've just been using it locally for the past bit and it appears to work great! 🙌

releasenotes/notes/fix-pythonpath-3e26d996b9df139a.yaml
P403n1x87: Great to hear that's working well! 😅  I've removed the relnote and moved the message to the commit 👍

riot/riot.py
brettlangdon: you are suggesting we do the opposite of what we have today, which is run everything in the base venv, and then add "child"/"other" site packages paths to add the necessary dependencies

compared to today, which we create the child venvs with the extra dependencies, and then add the parent venv site packages to PYTHONPATH to grab all the base dependencies + the library.



writing it out, it doesn't seem like there is actually that much of a difference between the two?

brettlangdon: nope, I just read the description for #143 and I was reading this comment wrong.

we create the base venv with base library, and then use pip prefix to just segment where we have pip install the unique/extra packages so we skip creating entire child venvs

that sounds good to me.

Kyle-Verhoog (approved):
👏

PR #143, opened 2021-07-16
perf: create base venvs only

P403n1x87:
With control over the pip installation prefix there is no need to create venvs other than the base one. This saves a lot of memory and speeds up the task runs.

These are some numbers based on the `riot` test suite when run with `riot run -p3.9 test`

## Before

`.riot` folder size: 184 MB
test run time: 63 s

## After

`.riot` folder size: 27 MB
test run time: 34 s

## Bonus

`riot list` works 🙂 

~~~
❯ pipenv run riot list     
test  Python Interpreter(_hint='3.6') 'pytest' 'pytest-cov' 'mock' 'typing-extensions'
test  Python Interpreter(_hint='3.7') 'pytest' 'pytest-cov' 'mock' 'typing-extensions'
test  Python Interpreter(_hint='3.8') 'pytest' 'pytest-cov' 'mock' 'typing-extensions'
test  Python Interpreter(_hint='3.9') 'pytest' 'pytest-cov' 'mock' 'typing-extensions'
black  Python Interpreter(_hint='3') 
fmt  Python Interpreter(_hint='3') 
flake8  Python Interpreter(_hint='3') 'flake8' 'flake8-blind-except' 'flake8-builtins' 'flake8-docstrings' 'flake8-import-order' 'flake8-logging-format' 'flake8-rst-docstrings' 'pygments'
mypy  Python Interpreter(_hint='3') 'mypy' 'pytest'
codecov  Python Interpreter(_hint='3') 'coverage'
docs  Python Interpreter(_hint='3') 'sphinx==3.3' 'sphinx-rtd-theme==0.5.0' 'sphinx-click==2.5.0' 'reno'
servedocs  Python Interpreter(_hint='3') 
releasenote  Python Interpreter(_hint='3') 
reno  Python Interpreter(_hint='3')
~~~

### DIFF
--- a/riot/riot.py
+++ b/riot/riot.py
@@ -246,31 +246,28 @@ class VenvInstance:
     parent: t.Optional["VenvInstance"] = None
 
     @property
-    def venv_path(self) -> t.Optional[str]:
-        """Return path to directory of the venv this instance should use.
+    def prefix(self) -> t.Optional[str]:
+        """Return path to directory where dependencies should be installed.
 
-        This will return a python version + package specific venv path name.
-        If no packages are defined it will return the ``Interpreter.venv_path``.
+        This will return a python version + package specific path name.
+        If no packages are defined it will return ``None``.
         """
-        if self.py is None:
+        if self.py is None or not self.pkgs:
             return None
 
         venv_path = self.py.venv_path
-        if self.needs_venv:
-            venv_path = "_".join(
-                [venv_path] + [f"{n}{rmchars('<=>.,', v)}" for n, v in self.pkgs]
-            )
-        return venv_path
+        assert venv_path is not None, self
+
+        ident = self.ident
+        assert ident is not None, self
+        return "_".join((venv_path, ident))
 
     @property
-    def target(self) -> t.Optional[str]:
-        """Target path for package installation."""
-        return self.venv_path
-
-    @property
-    def needs_venv(self) -> bool:
-        """Whether this ``VenvInstance`` needs its own venv or not."""
-        return bool(self.pkg_str)
+    def ident(self) -> t.Optional[str]:
+        """Return prefix identifier string based on packages."""
+        if not self.pkgs:
+            return None
+        return "_".join((f"{n}{rmchars('<=>.,', v)}" for n, v in self.pkgs))
 
     @property
     def pkg_str(self) -> str:
@@ -295,10 +292,10 @@ class VenvInstance:
 
     @property
     def bin_path(self) -> t.Optional[str]:
-        target = self.target
-        if target is None:
+        prefix = self.prefix
+        if prefix is None:
             return None
-        return os.path.join(target, "bin")
+        return os.path.join(prefix, "bin")
 
     @property
     def scriptpath(self):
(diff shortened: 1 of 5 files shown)


riot/riot.py
```diff
@@ -246,31 +246,28 @@ class VenvInstance:
...
 
     @property
-    def needs_venv(self) -> bool:
-        """Whether this ``VenvInstance`` needs its own venv or not."""
-        return bool(self.pkg_str)
+    def ident(self) -> t.Optional[str]:
+        """Return prefix identifier string based on packages."""
+        if not self.pkgs:
+            return None
+        return "_".join((f"{n}{rmchars('<=>.,', v)}" for n, v in self.pkgs))
```
brettlangdon, line 270: crazy idea... I wonder if we should normalize like `>=` to `gte_` so `flask>=1.0.0` -> `flask_gte_100`

also, are we going to have issues with `flask>=1.1.0` and `flask>=1.10.0` ?

(this is an aside, non-blocking for this PR

riot/riot.py
```diff
@@ -345,53 +342,44 @@ def pythonpath(self) -> str:
 
         return ":".join(paths)
 
+    def match_venv_pattern(self, pattern):
```
brettlangdon, line 345: docstring, maybe with some examples, might be nice

brettlangdon: tests?

brettlangdon (commented):
looks good, I like this change

riot/riot.py
P403n1x87: That's a good point. The more unique we make this the better! 🙂

Kyle-Verhoog: also `~`!

riot/riot.py
P403n1x87: This is tested with `test_list_with_venv_pattern` and `test_run_venv_pattern`, but I'm happy to add some unit tests if you think that's worth. What makes me a bit uneasy is that `venv_pattern` is no longer the right wording as we're not matching a venv anymore. Changing this would break the CLI though.

riot/riot.py
brettlangdon: `flask~=1.0.0` -> `flask_almost_100` 😆

riot/riot.py
brettlangdon: hmm, right

this is where maybe some unit tests would be good, we can enumerate all the possible/expected matching/patterns we support

brettlangdon: (or we can expand the others tests to cover them too)

P403n1x87:
There seems to be an issue with `mypy` when it is installed in its own prefix: https://github.com/python/mypy/issues/10829. We could wait for a resolution on the linked issue or temporarily take out `mypy` from the `riotfile.py` and run it the traditional way so that we can benefit from this optimisation.

Kyle-Verhoog:
Thanks for digging into it @P403n1x87! Yeah I'm ok with just running mypy outside of riot for now :+1:

brettlangdon (approved):
lgtm

brettlangdon (approved):

PR #145, opened 2021-07-20
fix: handling of package installs, env var overrides and namespace packages

P403n1x87:
With the idea of adding to the `PYTHONPATH`, each venv layer only needs to install its packages inside the venv instance, thus reducing the number of packages that are installed in each venv instance. We still use the full package spec in the result report for easier identification of all the dependencies used.

This change also fixes propagation and override of environment variables across layers as well as the correct handling of namespace packages.

BREAKING CHANGE: this change implies that package declarations that are used for pinning dependencies will have to be added to each venv layer that requires them.

### DIFF
--- a/riot/riot.py
+++ b/riot/riot.py
@@ -106,7 +106,7 @@ class Interpreter:
         return os.path.join(self.venv_path, "bin")
 
     @property
-    def site_packages_path(self) -> t.Optional[str]:
+    def site_packages_path(self) -> str:
         version = ".".join((str(_) for _ in self.version_info()[:2]))
         return os.path.join(self.venv_path, "lib", f"python{version}", "site-packages")
 
@@ -210,13 +210,18 @@ class Venv:
         parent_inst: t.Optional["VenvInstance"] = None,
     ) -> t.Generator["VenvInstance", None, None]:
         # Expand out the instances for the venv.
-        for env in expand_specs(self.env):
+        for env_spec in expand_specs(self.env):
+            # Bubble up env
+            env = parent_inst.env.copy() if parent_inst else {}
+            env.update(dict(env_spec))
+
             # Bubble up pys
             pys = (
                 self.pys
                 or (parent.pys if parent else None)
                 or [parent_inst.py if parent_inst else None]
             )
+
             for py in pys:
                 for pkgs in expand_specs(self.pkgs):
                     inst = VenvInstance(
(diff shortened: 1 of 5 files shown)


jd:
Do we need to change the riotfile in dd-trace-py? Should we prepare that change in parallel?

P403n1x87:
> Do we need to change the riotfile in dd-trace-py? Should we prepare that change in parallel?

Unfortunately we do. We need more changes like the one for boto :(. Doing this in parallel might be tricky as it's not totally obvious to me what will need to change.

Kyle-Verhoog (commented):
lgtm with a release note :+1:

riot/riot.py
```diff
@@ -556,19 +571,54 @@ def run(
...
                 )
+                # Copy over *nspkg.pth files so that we can handle namespace packages.
+                # This is strictly required by Python 2.7 and <3.3 (see PEP 420).
+                ns_files = []
                 try:
-                    # Pipe the command output directly to `out` since we
-                    # don't need to store it.
+                    sitepkgs = inst.site_packages_path
+                    assert sitepkgs is not None, inst
+                    for path in (_ for _ in inst.site_packages_list if _ != sitepkgs):
```
brettlangdon, line 587:
```suggestion
                    if inst.python_version < (3, 3):
                        for path in (_ for _ in inst.site_packages_list if _ != sitepkgs):
```

?

riot/riot.py
```diff
@@ -556,19 +571,54 @@ def run(
...
+                                    if os.path.isfile(ns_dest):
+                                        # child overrides parent
+                                        continue
+                                    ns_src = os.path.join(path, ns)
+                                    with open(ns_src) as ns_in:
+                                        # https://github.com/pypa/setuptools/blob/b62705a84ab599a2feff059ececd33800f364555/setuptools/namespaces.py#L44
+                                        content = ns_in.read().replace(
+                                            "sys._getframe(1).f_locals['sitedir']",
+                                            f"'{path}'",
+                                        )
```
brettlangdon, line 605: oh boy, this is gross

riot/riot.py
P403n1x87: That might work too, but from tests it doesn't look like it hurts doing the same thing for all Python versions so I'd keep the current code complexity. WDYT?

brettlangdon: yeah, figured if we can lock the workaround to older versions only then would reduce chances of bugs, at least for newer versions.

I don't have a really strong preference here, just a thought.

riot/riot.py
P403n1x87: very ☹️  but at least should be reliable 🤞  (and `sys._getframe(1).f_locals['sitedir']` isn't quite the nicest thing in the world either 😝 )

brettlangdon (approved):

PR #147, opened 2021-07-29
feat(shell): add riot shell

P403n1x87:
Screenshot 2021-10-12 at 11 11 49

### DIFF
--- a/riot/cli.py
+++ b/riot/cli.py
@@ -6,8 +6,12 @@ import sys
 
 import click
 import pkg_resources
+from rich.logging import RichHandler
 
 from .riot import Interpreter, Session
+
+FORMAT = "%(message)s"
+
 
 try:
     __version__ = pkg_resources.get_distribution("riot").version
@@ -55,7 +59,12 @@ PYTHON_VERSIONS_ARG = click.option(
 @click.pass_context
 def main(ctx, riotfile, log_level):
     if log_level:
-        logging.basicConfig(level=log_level)
+        logging.basicConfig(
+            level=log_level or logging.WARNING,
+            format=FORMAT,
+            datefmt="[%X]",
+            handlers=[RichHandler()],
+        )
 
     ctx.ensure_object(dict)
     try:
--- a/riot/cli.py
+++ b/riot/cli.py
@@ -4,12 +4,10 @@ import logging
 import re
 import sys
 
-
 import click
 import pkg_resources
 
 from .riot import Interpreter, Session
-
 
 try:
     __version__ = pkg_resources.get_distribution("riot").version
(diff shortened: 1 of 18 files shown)


riotfile.py
```diff
@@ -48,7 +48,7 @@
             name="mypy",
             command="mypy --install-types --non-interactive {cmdargs}",
             pkgs={
-                "mypy": latest,
+                "git+https://github.com/p403n1x87/mypy": "@fix-handle-prefix",
```
Kyle-Verhoog, line 51: this should be done separately?

tests/test_integration.py
```diff
@@ -506,9 +508,8 @@ def test_dev_install_fail(tmp_path: pathlib.Path, tmp_run: _T_TmpRun) -> None:
 """,
     )
     result = tmp_run("riot run test")
-    assert 'ERROR: File "setup.py"' in result.stderr
-    assert "Dev install failed, aborting!" in result.stderr
-    assert result.stdout == ""
+    assert 'File "setup.py"' in result.stderr
+    assert "Dev install failed, aborting!" in result.stdout
```
Kyle-Verhoog, line 512: 🤔 do we want this to go to stdout

Kyle-Verhoog (commented):
Awesome!! Just a couple things I saw but otherwise looks great

riotfile.py
P403n1x87: I'm not sure we'd really want to do this anyway. I've put it so that we can see the `mypy` check pass in CI. But if we think this is OK we should do it in a separate PR. At least one other change would have to be pulled out out here as it's not quite related, so a bit of clean-up is still needed.

riot/cli.py
```diff
@@ -53,13 +56,23 @@ def convert(self, value, param, ctx):
 )
 @click.option("-v", "--verbose", "log_level", flag_value=logging.INFO)
 @click.option("-d", "--debug", "log_level", flag_value=logging.DEBUG)
+@click.option("-P", "--pipe", "pipe_mode", is_flag=True, default=False)
```
Kyle-Verhoog, line 59: Is there a doc option? We should start documenting our flags/options to help usability

P403n1x87: Good point. It looks like `click.option` accepts a `help` argument. Given that this is not used with other options, perhaps we could leave this as a follow-up task?

Kyle-Verhoog: I think it makes sense to add it here still for `pipe` and then do the others as follow ups

Kyle-Verhoog (commented):
👍 looks good to me just the one help string 🙂

riot/riot.py
```diff
@@ -9,15 +10,38 @@
...
+from rich import print as rich_print
+from rich.pretty import Pretty
+from rich.status import Status
+from rich.table import Table
 
 logger = logging.getLogger(__name__)
 
 SHELL = os.getenv("SHELL", "/bin/bash")
 ENCODING = sys.getdefaultencoding()
+SHELL_RCFILE = """. ~/.bashrc
```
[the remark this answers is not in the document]
P403n1x87: > when I do `deactivate` it drops me into `bash`

That's because the new shell is a subprocess.

Kyle-Verhoog (approved):
🙌 🙌

PR #149, opened 2021-09-03
feat: add `create` argument to Venv

P403n1x87:
The new argument allows explicitly creating a virtual environment instance with all the dependencies installed within it rather than in separate prefixes. This is useful for packages that do not support dependencies installed in other prefixes.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "python.pythonPath": ".venv/bin/python3"
-}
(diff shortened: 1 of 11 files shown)


P403n1x87:
Requires #143.

.vscode/settings.json
```diff
@@ -1,3 +0,0 @@
-{
-    "python.pythonPath": ".venv/bin/python3"
```
P403n1x87, line 2: I think this file should be removed but I don't know how it got here in the first place 🤔

Kyle-Verhoog (approved):

PR #152, opened 2021-10-08
ci: add support for Python 3.10

P403n1x87:


jd (approved):

Kyle-Verhoog (dismissed):
Probably need to do something with cibuildwheels too

PR #157, opened 2021-10-12
Riot build base env option

mabdinur:


riot/riot.py
```diff
@@ -567,19 +569,22 @@ def run(
...
         results = []
 
-        self.generate_base_venvs(
-            pattern,
-            recreate=recreate_venvs,
-            skip_deps=skip_base_install,
-            pythons=pythons,
-        )
+        # if not installed or build option is set to true
+        if not self.are_base_venvs_generated or build:
```
brettlangdon, line 581: I think we need to actively check here if the base venv exists or not.

Since `are_base_venvs_generated` will only be True after `generate_base_venvs` is called, but we don't save this state between CLI runs.

We want to support workflows like:


```
$ riot generate  # base venv is created
$ riot run redis  # base venv is not created
```

Or

```
$ rm -rf .riot
$ riot run redis  # base venv gets created
$ riot run redis  # base venv does not get created
```

Or

```
$ riot generate  # base venv is created
$ riot run --build redis  # base venv is recreated
```

riot/cli.py
```diff
@@ -111,6 +111,7 @@ def generate(ctx, recreate_venvs, skip_base_install, pythons, pattern):
 @RECREATE_VENVS_ARG
 @SKIP_BASE_INSTALL_ARG
 @click.option("--pass-env", "pass_env", is_flag=True, default=False)
+@click.option("--build", "build", is_flag=True, default=False)
```
brettlangdon, line 114: I wonder what the end goal "action" is.

Like, are we "building" the dev package, are we "installing" it, are we "rebuilding"/"reinstalling" it? Or are we recreating the venv (e.g. delete it and create from scratch)?

e.g. I think any of these names could work (`--build` is acceptable! just brainstorming):

- `--build`
- `--install`
- `--recreate`

brettlangdon: - `--force-build`/`--force-install` ?

PR #162, opened 2022-01-25
feat: match pattern against VenvInstance name and short hash

### LINKED ISSUE #24: Improve test case selection mechanism
Right now CLI accepts a regex pattern of what to run, but it only allows you to filter on the suite name.

Instead we should allow people to:

* Run all suites
  * Default behavior?
* Run specific suites
  * e.g. `riot run black flake test`
* Run suites based on a matching pattern
  * e.g. "^django*" (to match "django", "djangorestframework", etc)
* Run a specific case
  * Run a suite with a specific set of dependencies selected, like Python 3.8 and Django 2.0.0
* Run cases based on a matching pattern
  * Either scoped to a specific suite, or across all suites
  * e.g. run all Python 3.8 tests across all suites

We should try to enumerate all the cases we have that we need to run before we try to design a CLI/API interface for it.

brettlangdon:
This allows you to run or list a specific venv instance.

`riot run <hash>`

Fixes  #24

### DIFF
--- a/releasenotes/notes/feat-match-short-hash-074e7595ec37967b.yaml
+++ b/releasenotes/notes/feat-match-short-hash-074e7595ec37967b.yaml
@@ -1,0 +1,5 @@
+---
+features:
+  - |
+    Match pattern against ``VenvInstance.name`` and ``VenvInstance.short_hash``.
+    This allows you to ``riot run <hash>`` to run only a specific instance.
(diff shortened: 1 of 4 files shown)


Kyle-Verhoog (dismissed):

jd (dismissed):
It seems quite easy to have some (unit) tests around this.

Kyle-Verhoog (approved):

PR #165, opened 2022-01-26
fix: make pip error message test more robust

P403n1x87:
The error message generated by pip when a package fails to install has changed, causing an integration test to fail. This change tries to make the test more future-proof.


brettlangdon (approved):

brettlangdon:
@Mergifyio refresh

PR #175, opened 2022-04-22
feat: use system default shell

P403n1x87:
The shell command has been changed to spawn the default shell instead of enforcing the use of bash.
Screenshot 2022-04-22 at 12 29 26


brettlangdon (approved):
💯 💯 💯 💯 💯

PR #178, opened 2022-05-27
refactor: make Venv a simple class

P403n1x87:
The release 0.960 of mypy has highlighted a misuse of `InitVar`. This change fixes it by removing the `dataclass` decorator from `Venv` and adding the required boilerplate code for the `__init__` method. The different `__repr__` method that follows from this refactor has caused the short hashes to change and the tests had to be adjusted accordingly.

### DIFF
--- a/riot/riot.py
+++ b/riot/riot.py
@@ -50,8 +50,8 @@ else:
     _T_CompletedProcess = subprocess.CompletedProcess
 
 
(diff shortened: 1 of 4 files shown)


Kyle-Verhoog (approved):
🙌 i definitely found the dataclasses 🪄 to be too much (and i think i wrote it lol 💀)

PR #179, opened 2022-07-19
Fix: Ensure a "restart" of virtualenv deleting the old one with recreation option (-r)

avara1986:
# Description
`riot` runs `virtualenv [directory]` to create a virtualenv. If `venv 01` directory exists and you run `virtualenv venv01`, the old virtualenv still has the same packages with no reset.

## Manually test
```
# Create a new virtualenv
python3.8 -m virtualenv -p38 venv_test_38
source venv_test_38/bin/activate 
pip install flask
deactivate

# Recreate
python3.8 -m virtualenv -p38 venv_test_38
source venv_test_38/bin/activate 
pip freeze
```
Output:

```
click==8.1.3
Flask==2.1.3
importlib-metadata==4.12.0
itsdangerous==2.1.2
Jinja2==3.1.2
MarkupSafe==2.1.1
Werkzeug==2.1.2
zipp==3.8.1
```

### DIFF
--- a/riot/riot.py
+++ b/riot/riot.py
@@ -169,9 +169,7 @@ class Interpreter:
             return venv_path
 
         if recreate:
-            logger.info(
-                "Deleting virtualenv '%s'", venv_path
-            )
+            logger.info("Deleting virtualenv '%s'", venv_path)
             shutil.rmtree(venv_path, ignore_errors=True)
 
         py_ex = self.path()
--- a/riot/riot.py
+++ b/riot/riot.py
@@ -168,6 +168,12 @@ class Interpreter:
             )
             return venv_path
 
+        if recreate:
+            logger.info(
+                "Deleting virtualenv '%s'", venv_path
+            )
+            shutil.rmtree(venv_path, ignore_errors=True)
+
         py_ex = self.path()
         logger.info("Creating virtualenv '%s' with interpreter '%s'.", venv_path, py_ex)
         run_cmd(["virtualenv", f"--python={py_ex}", venv_path], stdout=subprocess.PIPE)
--- a/riot/riot.py
+++ b/riot/riot.py
@@ -163,15 +163,14 @@ class Interpreter:
         """Attempt to create a virtual environment for this intepreter."""
         venv_path: str = path or self.venv_path
 
-        if os.path.isdir(venv_path) and not recreate:
-            logger.info(
-                "Skipping creation of virtualenv '%s' as it already exists.", venv_path
-            )
-            return venv_path
-
-        if recreate:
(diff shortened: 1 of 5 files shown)


tests/test_unit.py
```diff
@@ -129,3 +133,131 @@ def test_venv_name_matching(pattern: str) -> None:
     )
     assert venv.short_hash == "137331a"
     assert venv.matches_pattern(re.compile(pattern))
+
+
+# @pytest.fixture
```
brettlangdon, line 138: should these be deleted?

tests/test_cli.py
```diff
@@ -8,6 +8,7 @@
 import click.testing
 import mock
 import pytest
+
```
brettlangdon, line 11: unnecessary change in this PR?

although it seems like maybe isort/black should have been failing here 🤔 

but can we open as a separate PR if it does need to be fixed?

avara1986: Isort says that this space "should be there", I move it to an other PR :+1:

P403n1x87:
> @P403n1x87 you mean every time we do `riot run` it recreates the virtualenv and re-installs the dependencies?

Yes, to ensure that we test with exactly what we declare, and remove any potential mess created while messing around with `riot shell` inside the venvs. Currently, I'm doing `rm -rf .riot` as a last resort. I wonder if this also answers your question, @avara1986 ? I'm not sure I understand what `restart` means here. I think I might have missed some offline discussions?

avara1986:
@P403n1x87 as we talk, I added tests to check nested environments :muscle:

riot/riot.py
```diff
@@ -168,6 +169,10 @@ def create_venv(self, recreate: bool, path: t.Optional[str] = None) -> str:
             )
             return venv_path
 
+        if recreate:
```
P403n1x87, line 172: Since we are already testing for the directory on the previous `if`, perhaps we can rewrite that as

~~~
if os.path.isdir(venv_path)
    if recreate:
        ...
    else:
        ...
~~~
and remove the `ignore_errors=True`. WDYT?

avara1986: :+1:

avara1986: :+1:

tests/setup.py
```diff
@@ -0,0 +1,9 @@
+"""
+Setup file for testing purpose.
+
+Session.generate_base_venvs executes `pip ... install -e .` and
```
brettlangdon, line 4: Why was this not needed before?

P403n1x87: with #183 merged I think we don't need this anymore

P403n1x87 (approved):

PR #182, opened 2022-07-22
feat: change default riot path with a env var

avara1986:
# Features:
- Add `RIOT_ENV_BASE_PATH` environment variable to change the default path of `.riot` directory instead of current path.

# Motivation:
- It's needed in this PR #179 to create riot virtualenvs for testing purposes. 
- It's a useful feature, for example, set the .riot directory in `/tmp/riot_envs`.


riot/riot.py
```diff
@@ -156,7 +156,8 @@ def path(self) -> str:
     def venv_path(self) -> str:
         """Return the path to the virtual environment for this interpreter."""
         version = self.version().replace(".", "")
-        return os.path.abspath(f".riot/venv_py{version}")
+        env_base_path = os.environ.get("RIOT_ENV_BASE_PATH") or ".riot/venv_py"
```
brettlangdon, line 159: do we care to keep the `venv_py` prefix as well?

e.g. `RIOT_ENV_BASE_PATH` default is `.riot` only?

avara1986: My thoughts were "if you can edit the path, why not edit ALL THE PATH?" :sweat_smile: 

:+1: I change it

brettlangdon: yeah, maybe it is weird that we set the virtual env name prefix in the same setting?

avara1986: My ambition was my sin, you're right :upside_down_face: code updated :+1:

releasenotes/notes/feat-change-default-riot-path-with-a-env-var-12d2c59e74af9599.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    Add `RIOT_ENV_BASE_PATH` environment variable to change the default path of `.riot` directory instead of current
+    path.
```
P403n1x87, line 5: newline

P403n1x87 (commented):
Nit, otherwise LGTM!

releasenotes/notes/feat-change-default-riot-path-with-a-env-var-12d2c59e74af9599.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    Add `RIOT_ENV_BASE_PATH` environment variable to change the default path of `.riot` directory instead of current
```
P403n1x87, line 4:
```suggestion
    Add ``RIOT_ENV_BASE_PATH`` environment variable to change the default path of ``.riot`` directory instead of current
```
I think we need to double-backtick-quote this? I can never remember 😢 @brettlangdon @Kyle-Verhoog can you confirm?

ZStriker19: Yeah Gab you're correct, we need double backticks for envars.

riot/riot.py
```diff
@@ -165,7 +168,10 @@ def path(self) -> str:
     def venv_path(self) -> str:
         """Return the path to the virtual environment for this interpreter."""
         version = self.version().replace(".", "")
-        return os.path.abspath(f".riot/venv_py{version}")
+        env_base_path = os.environ.get("RIOT_ENV_BASE_PATH") or DEFAULT_RIOT_PATH
```
ZStriker19, line 171:
```suggestion
        env_base_path = os.environ.get("RIOT_ENV_BASE_PATH", DEFAULT_RIOT_PATH)
```
Pretty sure you can just specify the default like this.

brettlangdon (approved):

PR #187, opened 2022-08-17
feat: re-instal dev package on setup file changes

P403n1x87:
This change enables tracking changes in setup files so that the dev package can be re-installed if needed, when the -s option is passed to the run command.

### DIFF
--- a/riot/riot.py
+++ b/riot/riot.py
@@ -7,6 +7,7 @@ import importlib.util
 import itertools
 import logging
 import os
+from pathlib import Path
 import shutil
 import subprocess
 import sys
@@ -25,6 +26,7 @@ logger = logging.getLogger(__name__)
 
 DEFAULT_RIOT_PATH = ".riot"
 DEFAULT_RIOT_ENV_PREFIX = "venv_py"
+SETUP_FILES = ["setup.py", "pyproject.toml"]
 
 SHELL = os.getenv("SHELL", "/bin/bash")
 ENCODING = sys.getdefaultencoding()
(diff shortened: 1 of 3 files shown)


riot/riot.py
```diff
@@ -25,6 +26,7 @@
 
 DEFAULT_RIOT_PATH = ".riot"
 DEFAULT_RIOT_ENV_PREFIX = "venv_py"
+SETUP_FILES = ["setup.py", "pyproject.toml"]
```
jd, line 29: `setup.cfg`?

P403n1x87: Good question. I think `pip` looks for just `setup.py` and/or `pyproject.toml`, but `setup.cfg` should probably be hashed as well, if present 🤔

brettlangdon: do we want to add `setup.cfg` here? probably a good idea since it can contain install/build requirements

... would that mean we also care about `requirements.txt`? or any variation of those?

brettlangdon (commented):
should we just fail instead if `-s` is passed and we noticed a change?

what about c-extensions? we would definitely want to rebuild in those cases?

would this become any easier if we removed `-s` and made that the default behavior? e.g. `riot run <venv>` would only build the base venv if it doesn't exist, otherwise you use `riot generate` or can pass in an explicit option like `riot run --rebuild`?

riot/riot.py
mabdinur, line 29:
```suggestion
SETUP_FILES = ["setup.py", "pyproject.toml", "setup.cfg"]
```

PR #191, opened 2022-10-06
fix: Update interpreters option for riot list to sort versions and output format

Yun-Kim:
The `riot list -i` flag output was updated to have a newline-separated format instead of comma-separated, and to be sorted by version for consistency.


tests/data/diff_pys_riotfile.py
```diff
@@ -0,0 +1,14 @@
+from riot import Venv
+
+venv = Venv(
+    venvs=[
+        Venv(
+            name="test",
+            command="exit 0",
+            pys=["2.7", "3.5", "3.6", "3.7", "3.8", "3.9"],
```
brettlangdon, line 8: can we make this unsorted to verify that it does get sorted?

Yun-Kim: Thanks for the good catch!

tests/data/diff_pys_riotfile.py
```diff
@@ -0,0 +1,14 @@
+from riot import Venv
+
+venv = Venv(
+    venvs=[
+        Venv(
+            name="test",
+            command="exit 0",
+            pys=["3.5", "3.8", "3.6", "3.9", "3.7", "2.7"],
```
brettlangdon, line 8:
```suggestion
            # DEV: purposely out of order so we can verify sorting in tests
            pys=["3.5", "3.8", "3.6", "3.9", "3.7", "2.7"],
```



?

brettlangdon (approved):

riot/riot.py
```diff
@@ -7,6 +7,8 @@
 import itertools
 import logging
 import os
+from packaging.version import parse
```
brettlangdon, line 10: is this used?

brettlangdon (approved):

PR #194, opened 2022-11-16
feat(venv): add pip install options

mabdinur:
Adds support for passing [pip install options](https://pip.pypa.io/en/stable/cli/pip_install/#pip-install) to riot venvs.

### DIFF
--- a/releasenotes/notes/add-pip-install-options-1d99d61f24e8ee15.yaml
+++ b/releasenotes/notes/add-pip-install-options-1d99d61f24e8ee15.yaml
@@ -1,0 +1,4 @@
+---
+features:
+  - |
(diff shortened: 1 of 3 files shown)


releasenotes/notes/add-pip-install-options-1d99d61f24e8ee15.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    adds `install_options` field to `riot.Venv`. This field can be used to pass options to `pip install` commands.
```
brettlangdon, line 4: we should be able to handle these via `env` as well, https://pip.pypa.io/en/stable/topics/configuration/#environment-variables

mabdinur:
This feature is not required. https://pip.pypa.io/en/stable/topics/configuration/#environment-variables can be used instead.

PR #196, opened 2022-11-18
feat: add requirements caching

majorgreys:
This change adds the `riot requirements <hash>` command, which uses pip-compile to generate a requirements lockfile for the venv identified by the hash. It also adds the `-c/--recompile-requirements` flag to `riot run`. When set, this flag causes the requirements.txt lockfile and the venv itself to be regenerated. When omitted, the requirements lockfile is only built if it does not already exist.

This is a useful feature because it reduces duplicated work during `riot run`. In current master, every `riot run` has to build the dependency tree for its venv before installing those dependencies. With this change, the built dependency tree is saved in the form of a lockfile and reused on subsequent `riot run` invocations.

### DIFF
--- a/riot/riot.py
+++ b/riot/riot.py
@@ -198,7 +198,10 @@ class Interpreter:
 
         py_ex = self.path()
         logger.info("Creating virtualenv '%s' with interpreter '%s'.", venv_path, py_ex)
-        run_cmd(["virtualenv", f"--python={py_ex}", venv_path], stdout=subprocess.PIPE)
+        run_cmd(
+            ["virtualenv", f"--python={py_ex}", venv_path],
+            stdout=subprocess.PIPE,
+        )
 
         return True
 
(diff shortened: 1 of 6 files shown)


riot/riot.py
```diff
@@ -457,6 +460,41 @@ def __hash__(self):
...
+    @property
+    def requirements(self) -> str:
+        """Requirements for dependencies with pinned versions."""
+        # Transform full_pkg_str into requirements.in format
+        pkgs = "\n".join(self.full_pkg_str.replace("'", "").split(" "))
+        _dir = os.path.join(DEFAULT_RIOT_PATH, "requirements")
+        os.makedirs(_dir, exist_ok=True)
+        in_path = os.path.join(_dir, "{}.in".format(self.short_hash))
+        subprocess.check_output(
+            [self.py.path(), "-m", "pip", "install", "pip-tools"],
```
brettlangdon, line 472: do we just need to add `pip-tools` as a default requirement in `setup.py`?

brettlangdon: or does it have to be in each venv?

riot/riot.py
```diff
@@ -557,15 +598,24 @@ def prepare(
...
             assert pkg_str is not None
+            compiled_requirements_file = (
+                f"{DEFAULT_RIOT_PATH}/requirements/{self.short_hash}.txt"
+            )
+            if recompile_reqs or not os.path.exists(compiled_requirements_file):
+                _ = self.requirements
+            cmd = (
+                f"pip --disable-pip-version-check install --prefix '{self.prefix}' --no-warn-script-location "
+                f"-r {compiled_requirements_file}"
+            )
```
brettlangdon, line 609: I am not sure I follow this correctly.

We always require the requirements file?

riot/riot.py
emmettbutler: it has to be in each venv because it matters which version of python was used to install it.

riot/riot.py
emmettbutler: That's right. The .txt lockfile is always part of the `run` command. If it doesn't exist it's created, but if it does it's used and the work of creating it is skipped.

brettlangdon: oh! the `_ = self.requirements` is the part that threw me off

I wonder if we should not have that as a property but a function to be more explicit about it. wdyt?

tests/test_integration.py
```diff
@@ -896,3 +898,65 @@ def test_long_path(tmp_path: pathlib.Path, tmp_run: _T_TmpRun) -> None:
...
+
+    lockfile_path = str(
+        tmp_path
+        / os.path.join(
+            ".riot",
+            "requirements",
+            "{}.txt".format(_hash),
+        )
+    )
+    assert os.path.exists(lockfile_path)
```
brettlangdon, line 931: do we want to validate the contents?

we should trust `pip-tools`, but it should also be following a fairly expected format.

(we might need to pin specific pkg versions to get consistent behavior though, unless we just validate `requests==.*?$`)

brettlangdon (commented):
two considerations, but otherwise lgtm to start trying out!

brettlangdon (dismissed):

emmettbutler:
@brettlangdon I signed the commit to make CI pass, but the diff is the same. Could you re-approve?

brettlangdon (approved):

PR #199, opened 2023-01-09
fix: cap long prefix paths

### LINKED ISSUE #198: riot generates paths longer than the FS can support
If a virtual environment has many dependencies (the odds for "fat" venvs created with `create=True` are in fact higher), then the generated path could be longer than the FS can support and the venv creation fails. In such cases it's perhaps best replacing the path containing the package spec with the venv hash.

P403n1x87:
This change ensures that riot does not generate prefix paths that are too long and might go beyond the limits imposed by the FS.

Fixes #198.

### DIFF
--- a/tests/test_unit.py
+++ b/tests/test_unit.py
@@ -9,7 +9,6 @@ import pytest
 from riot.riot import Interpreter, run_cmd, Session, Venv, VenvInstance
 from tests.test_cli import DATA_DIR
 
-
 RIOT_TESTS_PATH = os.path.join(os.path.dirname(__file__), ".riot")
 default_venv_pattern = re.compile(r".*")
 current_py_hint = "%s.%s" % (sys.version_info.major, sys.version_info.minor)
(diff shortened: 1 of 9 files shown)


majorgreys:
Is it necessary to drop 3.6 support here?

P403n1x87:
> Is it necessary to drop 3.6 support here?

Not here, I can make another PR for that, but CI couldn't pull 3.6 anymore and I thought it didn't make sense to invest time in trying to fix it for such an old version.

ZStriker19:
What's FS?

ZStriker19 (approved):

mabdinur (approved):

mabdinur:
macos-latest 3.6 and ubuntu-latest 3.6 are required actions. Can we remove this check?

P403n1x87:
> What's FS?

File System 🙂

P403n1x87:
> macos-latest 3.6 and ubuntu-latest 3.6 are required actions. Should we remove this check to unblock this PR?

Yes, I think we should. But I would like to drop support for 3.6 in another PR first. I'll try to open it today 🤞

tests/test_unit.py
```diff
@@ -9,7 +9,6 @@
 from riot.riot import Interpreter, run_cmd, Session, Venv, VenvInstance
 from tests.test_cli import DATA_DIR
```
mabdinur, line 11:
```suggestion


```

mabdinur (approved):

ZStriker19 (approved):

PR #203, opened 2023-03-15
fix: change how Venvs are hashed

emmettbutler:
This change removes `self` from the seed used to generate Venv hashes. This is because recent changes to the requirements installation logic mutated `self` between installation and pythonpath generation in such a way as to change the resulting hash. This meant that the virtualenv pointed to in the pythonpath was not the same one that had been generated for the current command run.

Basing the hash solely on attributes that don't change during execution keeps it stable per run.

This has been tested on dd-trace-py https://github.com/DataDog/dd-trace-py/pull/5227/commits/492fc0d8b21e537541befce36ac77478c6ed4f29

### DIFF
--- a/releasenotes/notes/stable-hash-63389fa04cca60ab.yaml
+++ b/releasenotes/notes/stable-hash-63389fa04cca60ab.yaml
@@ -2,4 +2,4 @@
 fixes:
   - |
     Fixes a bug causing virtualenv hashes to change during command execution, leading to installed
-    dependencies not being found
+    dependencies not being found.
--- a/releasenotes/notes/stable-hash-63389fa04cca60ab.yaml
+++ b/releasenotes/notes/stable-hash-63389fa04cca60ab.yaml
@@ -1,0 +1,5 @@
+---
+fixes:
+  - |
+    Fixes a bug causing virtualenv hashes to change during command execution, leading to installed
+    dependencies not being found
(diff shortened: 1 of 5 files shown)


releasenotes/notes/stable-hash-63389fa04cca60ab.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    Fixes a bug causing virtualenv hashes to change during command execution, leading to installed
+    dependencies not being found
```
ZStriker19, line 5:
```suggestion
    dependencies not being found.
```

ZStriker19: actually, does this even need a release note? I'm leaning towards no.

emmettbutler: I think it does because the bug this fixes is fairly serious.

emmettbutler: I thought a bit more about this and decided that it makes sense not to include a release note.

ZStriker19 (dismissed):

emmettbutler:
@ZStriker19 I had to dismiss your review to sign all of the commits on this branch. If you could re-approve when you get a moment, I'd appreciate it!

ZStriker19 (approved):

PR #205, opened 2023-04-20
fix: pin virtualenv version to 20.20.0

gnufede:
Virtualenv dropped support for Python 2.7, so we need to pin this requirement in order to keep creating 2.7 virtualenvs

### DIFF
--- a/releasenotes/notes/fix-python-2-7-pinning-virtualenv-619b531a9ea8c695.yaml
+++ b/releasenotes/notes/fix-python-2-7-pinning-virtualenv-619b531a9ea8c695.yaml
@@ -1,0 +1,4 @@
+---
+fixes:
+  - |
+    Avoid using latests versions of ``virtualenv`` since it has dropped support for Python 2.7.
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ setup(
     install_requires=[
         "dataclasses; python_version<'3.7'",
         "click>=7",
-        "virtualenv==20.20.0",
+        "virtualenv<=20.20.0",
         "rich",
         "pexpect",
         "packaging",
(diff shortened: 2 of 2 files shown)


avara1986 (approved):

brettlangdon (changes requested):
we just need a changelog entry, otherwise lgtm

P403n1x87:
Awesome find! Should we perhaps make this more general and pin _all_ dependencies?

brettlangdon:
> Awesome find! Should we perhaps make this more general and pin _all_ dependencies?

I'd prefer we not, same as our other packages, we should only set minimum required versions....

in fact for this, we could probably just do `virtualenv<=20.20.0` to avoid upgrading someones own version of virtualenv.

P403n1x87:
IMO we should install riot in its isolated environment as a stand-alone tool (e.g. with `pipx`) and pin all dependencies. I don't think that we would ever have the need to "inject" riot in other environments.

gnufede:
> we just need a changelog entry, otherwise lgtm

Should it be a `fix` entry? or what kind?

brettlangdon:
> IMO we should install riot in its isolated environment as a stand-alone tool (e.g. with `pipx`) and pin all dependencies. I don't think that we would ever have the need to "inject" riot in other environments.

We aren't the only users of `riot` though. so instead would be better to be as open/permissive on dependencies as possible.

If we want we can/should have a pipx/pip-compile lock file for riot where we install it in our CI environment?

brettlangdon (approved):

brettlangdon:
@P403n1x87 we can probably handle as a follow-up, since this fix vs changing how riot defines dependencies can be separated. wdyt?

P403n1x87:
> @P403n1x87 we can probably handle as a follow-up, since this fix vs changing how riot defines dependencies can be separated. wdyt?

👍  yep, that's fine by me.

PR #209, opened 2023-05-24
fix: install dev package in its own prefix

P403n1x87:
Resolves: https://github.com/DataDog/riot/issues/211

This prevents any dependency layers from removing the dev package dependencies from the base environment.

### DIFF
--- a/releasenotes/notes/fix-dev-pkg-prefix-cdce8c029dfe3dff.yaml
+++ b/releasenotes/notes/fix-dev-pkg-prefix-cdce8c029dfe3dff.yaml
@@ -1,0 +1,5 @@
+---
+fixes:
+  - |
+    Fixed a problem that caused dependency layers to uninstall some dev package
+    dependencies from the base virtual environment.
--- a/riot/_hotfix/sitecustomize.py
+++ b/riot/_hotfix/sitecustomize.py
@@ -1,0 +1,22 @@
+import os
+
+from pip._internal.build_env import BuildEnvironment as BE
+
+original_be_enter = BE.__enter__
+
+
+def be_enter(self):
+    pythonpath = os.getenv("PYTHONPATH")
+    try:
+        return original_be_enter(self)
+    finally:
+        # We fix the PYTHONPATH override done by pip before returning
+        if pythonpath is not None:
+            os.environ["PYTHONPATH"] = os.pathsep.join(
+                (os.getenv("PYTHONPATH"), pythonpath)
+            )
+
+
+# pip does not support edit install with prefix in Python 2
+# https://github.com/pypa/pip/issues/7627
+BE.__enter__ = be_enter
--- a/riot/riot.py
+++ b/riot/riot.py
@@ -522,7 +522,11 @@ class VenvInstance:
             if self.py.bin_path is not None:
                 paths.append(self.py.bin_path)
 
-        return ":".join(paths)
+        dev_pkg_path = "_".join((self.py.venv_path, "dev_pkg"))
+        if os.path.exists(dev_pkg_path):
+            paths.append(os.path.join(dev_pkg_path, "bin"))
+
+        return os.pathsep.join(paths)
 
     @property
     def site_packages_path(self) -> t.Optional[str]:
@@ -611,7 +615,7 @@ class VenvInstance:
             if self.created:
                 py.create_venv(recreate, venv_path)
                 if not skip_deps:
-                    install_dev_pkg(venv_path, self.py.version_info())
+                    install_dev_pkg(venv_path, self.py.version_info(), in_prefix=False)
 
             pkg_str = self.pkg_str
             assert pkg_str is not None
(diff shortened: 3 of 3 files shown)


avara1986 (dismissed):

mabdinur:
Something interesting I noticed is that pip logs the following error in cherrypy when riot venvs are created ([link](https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/36358/workflows/605fdf92-445e-476d-9019-5043358e967a/jobs/2453524))

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
ddtrace 1.15.0.dev17+g5f8107360 requires typing-extensions, which is not installed
```

Is this expected? Would this fix also address this error?

P403n1x87:
> Something interesting I noticed is that pip logs the following error in cherrypy when riot venvs are created ([link](https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/36358/workflows/605fdf92-445e-476d-9019-5043358e967a/jobs/2453524))
> 
> ```
> ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
> ddtrace 1.15.0.dev17+g5f8107360 requires typing-extensions, which is not installed
> ```
> 
> Is this expected? Would this fix also address this error?

It looks like the dev package is not being installed. That's when the dependencies are installed with the proposed riot change. I wonder if there are some changes still needed to support the lockfiles, or if they have to be re-generated. Tagging @emmettbutler as I'm not that familiar with how that part of riot works.

mabdinur:
With this fix we are able to build base envs in ci: https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/36413/workflows/5058b2b6-b4e1-45f1-8b34-531217eb88ac/jobs/2456864, however it causes a different set of issues in ci: https://app.circleci.com/pipelines/github/DataDog/dd-trace-py?branch=munir%2Funblock-ci

P403n1x87:
Ah! Having moved the library to a different prefix, we also need to update the `PATH` to allow it to find `ddtrace-run` 😢 . I'll look into that tomorrow.

mabdinur:
Hmmm we still experiencing issues installing packages: [import error]( https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/36567/workflows/373b88fe-aef1-4481-bd19-aa20c6a021c5/jobs/2466295), [PR](https://github.com/DataDog/dd-trace-py/pull/5972). Ideally these issues should be caught by riot tests. Can we add a regression test?

Also this hot fix has a lot complexity and could be difficult to debug/maintain. Can we simplify it a bit 😅

P403n1x87:
> Hmmm we still experiencing issues installing packages: [import error](https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/36567/workflows/373b88fe-aef1-4481-bd19-aa20c6a021c5/jobs/2466295), [PR](https://github.com/DataDog/dd-trace-py/pull/5972). Ideally these issues should be caught by riot tests. Can we add a regression test?

Ah I think the slotscheck issue is to be expected since the dev package is no longer installed in the base venv. We would have to check whether the new site-packages is ending up in the PYTHONPATH. If not we need to somehow add it, but I'm not sure this is something to address in riot.

> Also this hot fix has a lot complexity and could be difficult to debug/maintain. Can we simplify it a bit 😅

I'm afraid I don't have anything simpler to propose 🙁

avara1986 (approved):

mabdinur:
I skipped the slotcheck

> > Hmmm we still experiencing issues installing packages: [import error](https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/36567/workflows/373b88fe-aef1-4481-bd19-aa20c6a021c5/jobs/2466295), [PR](https://github.com/DataDog/dd-trace-py/pull/5972). Ideally these issues should be caught by riot tests. Can we add a regression test?
> 
> Ah I think the slotscheck issue is to be expected since the dev package is no longer installed in the base venv. We would have to check whether the new site-packages is ending up in the PYTHONPATH. If not we need to somehow add it, but I'm not sure this is something to address in riot.
> 
> > Also this hot fix has a lot complexity and could be difficult to debug/maintain. Can we simplify it a bit 😅
> 
> I'm afraid I don't have anything simpler to propose 🙁

I skipped the slotscheck in this [PR](https://github.com/DataDog/dd-trace-py/pull/5972/files) and now most test runs are failing due import errors: https://app.circleci.com/pipelines/github/DataDog/dd-trace-py/36927/workflows/209ef960-c85f-4763-9ce1-727e5c1bdc92.  I am not sure if this new issue should be fixed in riot or if we need to update the `dd-trace-py` tests? I don't think we can move forward with this change without addressing this issue. 

@P403n1x87 any thoughts on what we should try next?

P403n1x87:
🤔  it looks like there are different kinds of failures. The ones failing to import `envier` seem to be missing the `dev_pkg` on the `PYTHONPATH` for some reason (and this is probably a bug in this change, but not obvious to me currently). The other failures (e.g. `debugger`) seem to be related to native extensions not being found. I wonder if we should do something more/different to ensure we can also find those native extensions.

P403n1x87:
Superseded by #212

PR #212, opened 2023-07-06
fix: use dedicated venv for installing dependencies

P403n1x87:
We let riot create an auxiliary venv, suffixed with `_deps`, that is used exclusively for installing dependencies in venv layers. This way we can keep the dev package intact in the base venv.

### DIFF
--- a/riot/riot.py
+++ b/riot/riot.py
@@ -134,9 +134,15 @@ class Interpreter:
         return os.path.join(self.venv_path, "bin")
 
     @property
+    def dev_pkg_bin_path(self) -> t.Optional[str]:
+        return os.path.join(self.dev_pkg_venv_path, "bin")
+
+    @property
     def site_packages_path(self) -> str:
         version = ".".join((str(_) for _ in self.version_info()[:2]))
-        return os.path.join(self.venv_path, "lib", f"python{version}", "site-packages")
+        return os.path.join(
+            self.dev_pkg_venv_path, "lib", f"python{version}", "site-packages"
+        )
 
     @functools.lru_cache()
     def path(self) -> str:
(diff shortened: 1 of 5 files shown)


riot/riot.py
```diff
@@ -1081,6 +1084,19 @@ def run_cmd_venv(
...
+        try:
+            sitepkgs_path = (
+                next((Path(abs_venv) / "lib").glob("python*")) / "site-packages"
+            )
+            pythonpath = env.get("PYTHONPATH", None)
+            env["PYTHONPATH"] = (
+                os.pathsep.join((pythonpath, str(sitepkgs_path)))
+                if pythonpath is not None
+                else str(sitepkgs_path)
+            )
```
mabdinur, line 1096: Can we add a comment here to call out why this is necessary

riot/riot.py
```diff
@@ -1081,6 +1084,19 @@ def run_cmd_venv(
         env["VIRTUAL_ENV"] = abs_venv
         env["PATH"] = f"{abs_venv}/bin:" + env.get("PATH", "")
 
+        try:
+            sitepkgs_path = (
+                next((Path(abs_venv) / "lib").glob("python*")) / "site-packages"
```
mabdinur, line 1089: Why do we need to call `next` here. Can we just use the first match?

P403n1x87: Here `next` is indeed getting the first (and hopefully only!) result from the iterator returned by `glob`

tests/test_integration.py
```diff
@@ -611,22 +611,17 @@ def test_interpreter_pythonpath(tmp_path: pathlib.Path, tmp_run: _T_TmpRun) -> N
...
+        tmp_path
+        / os.path.join(
+            ".riot",
+            f"venv_py{version}",
+            "lib",
+            f"python{py_dot_version}",
+            "site-packages",
         )
     )
+    assert env["PYTHONPATH"] == ":".join(("", str(tmp_path), sitepkgs, sitepkgs))
```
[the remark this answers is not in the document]
P403n1x87: Answered offline: the duplication is caused by the extra entry in PYTHONPATH added out of precaution by this change in `run_cmd_venv`. We could make the logic smarter and catch any duplication, but for now we can keep this nice and easy to get a fix out ASAP. We can then revisit this in a follow-up PR.

mabdinur (approved):

PR #220, opened 2023-08-09
feat: support 3.12 and drop support for 2.7 virtual environments

majorgreys:


brettlangdon (commented):
major version bump?

Yun-Kim (approved):

Yun-Kim:
@brettlangdon since riot is still in beta, I don't think we need to use this to warrant releasing a major version. I'll instead yank 0.18.0 since it breaks with Python 2.7 I believe and release this in 0.19.0.

PR #222, opened 2023-10-02
Move environment variable logging from verbose to debug log level

romainkomorn-exdatadog:
Logging environment variables risks leaking secrets, and likely isn't necessary in non-debugging scenarios.

Another, more complicated option, would be to have an allowlist (or blocklist) of environment variables to suppress from logging.


romainkomorndatadog:
Maybe this should've broken some tests. 🤔

PR #223, opened 2023-10-03
fix error out if list or run find no matching environments to run

romainkomorndatadog:
This updates the CLI to exit with error code 1 for the `list` and `run` commands if there ends up being no matches.

The intention here is to prevent scenarios where `list` or `run` end up matching nothing, giving the caller (or in the case motivating this PR: CI runs) the false impression that tests succeeded when, in fact, it is more likely that the pattern passed to `riot` wasn't properly formed and no tests were run at all.


P403n1x87 (approved):

romainkomorndatadog:
I'm going to close this in favor of "we're moving to hatch" context.

PR #239, opened 2026-01-09
feat: add support for installing pre-built wheels

brettlangdon:
This PR adds support for installing the project under test from pre-built wheels instead of editable mode. This is useful for testing with production-like builds, validating wheel packaging, and improving test reproducibility in CI/CD environments.

The feature is available via the `--wheel-path` global option or `RIOT_WHEEL_PATH` environment variable, which accepts either a local directory path or a remote URL. When specified, riot uses a two-step installation process (`pip download --no-index` + `pip install`) to ensure only wheels from the specified source are used, with no fallback to PyPI. Package names are automatically resolved from `pyproject.toml` or the `RIOT_PACKAGE_NAME` environment variable.

**Example usage:**
```bash
pip wheel --no-deps -w dist/ .
riot --wheel-path dist/ run test
```


emmettbutler (dismissed):

gnufede (dismissed):

emmettbutler (dismissed):
Looks great to me. I added the ability to install dependencies from local wheels, in addition to the "main" library. This facilitates the modularization work I'm doing in https://github.com/DataDog/dd-trace-py/pull/17991.

Partially tested via https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-py/-/jobs/1701268103 on https://github.com/DataDog/dd-trace-py/pull/17991. The following log from that job is a bit questionable, but that's my only concern:
```
ERROR: Invalid wheel filename (wrong number of parts): '*'
```

riot/runner.py
```diff
@@ -99,14 +102,93 @@ def install_dev_pkg(venv_path: str, force: bool = False) -> None:
...
-        dev_pkg_lockfile.touch()
-    except CmdFailure as e:
-        logger.error("Dev install failed, aborting!\n%s", e.proc.stdout)
-        sys.exit(1)
+
+        # Step 1: Download wheel to temp directory using --no-index to avoid PyPI
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            download_cmd = (
+                f"pip --disable-pip-version-check download "
+                f"--no-index --no-deps --find-links '{wheel_path}' "
```
brettlangdon:
```suggestion
                f"--no-index --no-deps {find_links_flag} "
```

no a blocker

emmettbutler: oversight, thanks

emmettbutler (approved):

PR #262, opened 2026-04-09
perf: cache full_pkg_str/long_hash and skip pip-tools reinstall

P403n1x87:
- Cache full_pkg_str and long_hash with cached_property so the parent chain traversal and SHA256 computation happen at most once per VenvInstance, regardless of how many times they are accessed.
- Fix the double-insert bug in full_pkg_str (self was added to the chain twice), using append + reversed for O(n) traversal.
- Skip pip-tools reinstall on every requirements compilation by writing a sentinel file (.riot/pip-tools-<venv>.installed) on first install. Subsequent compilations check for the file instead of spawning a subprocess.

### DIFF
--- a/riot/venv.py
+++ b/riot/venv.py
@@ -1,5 +1,6 @@
 from contextlib import contextmanager
 import dataclasses
+import functools
 from hashlib import sha256
 import logging
 import os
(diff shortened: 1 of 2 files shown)


emmettbutler (approved):
very nice!

PR #266, opened 2026-05-11
feat: add support for free-threaded mode

P403n1x87:
We add support for free-threaded mode by handling virtual environments created with a free-threaded build of CPython.

### DIFF
--- a/riot/interpreter.py
+++ b/riot/interpreter.py
@@ -29,17 +29,22 @@ class Interpreter:
         return repr(self)
 
     @functools.lru_cache()
-    def is_freethreaded(self) -> bool:
-        """Return whether this interpreter is a free-threaded (no-GIL) build."""
-        path = self.path()
+    def abiflags(self) -> str:
+        """Return the ABI flags for this interpreter (e.g. '' or 't' for free-threaded).
+
+        On POSIX, delegates to sys.abiflags. On Windows (where sys.abiflags does not
+        exist) falls back to sysconfig.get_config_var('Py_GIL_DISABLED') so that
+        free-threaded builds are still distinguished from regular ones.
+        """
         output = subprocess.check_output(
             [
-                path,
+                self.path(),
                 "-c",
-                "import sys; print(getattr(sys, '_is_gil_enabled', lambda: True)() is False)",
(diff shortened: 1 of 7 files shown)


riot/interpreter.py
```diff
@@ -28,6 +28,19 @@ def __str__(self) -> str:
...
 
+    @functools.lru_cache()
+    def is_freethreaded(self) -> bool:
+        """Return whether this interpreter is a free-threaded (no-GIL) build."""
+        path = self.path()
+        output = subprocess.check_output(
+            [
+                path,
+                "-c",
+                "import sys; print(getattr(sys, '_is_gil_enabled', lambda: True)() is False)",
```
gnufede: Can't we do this in the same process instead of starting a subprocess?

P403n1x87: we could be using e.g. Python 3.10 to run riot against a 3.13t venv, so we need to run the target interpreter in this case

brettlangdon:
```suggestion
                "import sysconfig; print(sysconfig.get_config_var('Py_GIL_DISABLED') == 1)",
```

should also work

there is also `"abiflags": "t"`

brettlangdon: actually, if we use `abiflags`, we probably don't need to care about free-threaded vs not

e.g. 

> suffix = "t" if self.is_freethreaded() else ""

just do:

> f"{DEFAULT_RIOT_ENV_PREFIX}{version}{self.abiflags}"

or something

P403n1x87: 💡

gnufede (dismissed):

riot/interpreter.py
P403n1x87: one note is that it's UNIX only 🤔

brettlangdon: yeah, but we don't run riot with windows do we? we don't test against it, afaik

brettlangdon (approved):

riot/interpreter.py
P403n1x87: correct, but I believe we want to run riot with dd-trace-py on Windows as well? 🤷

brettlangdon: I don't think we do today at all, we only run smoke tests on windows without riot, but being able to in the future, sure

PR #267, opened 2026-05-20
feat(requirements): add uv pip compile backend with --exclude-newer support

juanjux:
## Summary

Adds an opt-in `uv pip compile` backend to `VenvInstance.requirements`, selectable via two new environment variables:

- `RIOT_PIP_COMPILE_BACKEND`: `piptools` (default) or `uv`.
- `RIOT_PIP_COMPILE_EXCLUDE_NEWER`: forwarded to `uv pip compile` as `--exclude-newer=<value>`. Ignored (with a warning) when the backend is `piptools` because pip-tools has no equivalent option.

The historical behaviour is preserved exactly when neither variable is set. When the `uv` backend is selected the `uv` executable must already be on `PATH`; a clear `RuntimeError` is raised if it isn't.

## Why

Cross-language supply-chain hardening work in `dd-trace-py` (tracked in [DataDog/dd-trace-py#18182](https://github.com/DataDog/dd-trace-py/pull/18182)) needs a 48h "cooldown" on every release that ends up in a compiled lockfile, including transitive dependencies. `--exclude-newer=<date>` is the ergonomic primitive for that on `uv pip compile`; `pip-tools` has nothing equivalent. Rather than fork riot or post-process its output, projects can now flip a single env var and get the cooldown for free.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -28,3 +28,35 @@ Venv
 
 .. autoclass:: riot.venv.Venv
    :members:
+
+
+Environment variables
+---------------------
+
+The following environment variables influence riot's behaviour:
+
+``RIOT_PATTERN``
+    Default value for the ``pattern`` argument of ``list``, ``run`` and
+    ``generate``.
+
+``RIOT_ENV_BASE_PATH``
+    Base directory under which virtual environments are created.
+    Defaults to ``.riot`` (relative to the current working directory).
+
+``RIOT_PIP_COMPILE_BACKEND``
+    Selects the compiler used when generating lockfiles via the
+    ``requirements`` command. Supported values:
+
+    - ``piptools`` (default) — invoke ``python -m piptools compile``;
+      ``pip-tools`` is auto-installed on demand.
+    - ``uv`` — invoke ``uv pip compile``. The ``uv`` executable must
+      already be available on ``PATH``; a clear error is raised
+      otherwise.
+
+``RIOT_PIP_COMPILE_EXCLUDE_NEWER``
+    When set together with ``RIOT_PIP_COMPILE_BACKEND=uv``, the value is
+    forwarded to ``uv pip compile`` as ``--exclude-newer=<value>`` so
+    that releases published after the given timestamp are not picked up
+    during lockfile resolution. Useful for enforcing a supply-chain
+    "cooldown" on freshly published transitive dependencies. Ignored
+    with a warning when the backend is ``piptools``.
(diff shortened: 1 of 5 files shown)


docs/configuration.rst
```diff
@@ -28,3 +28,35 @@ Venv
...
+
+``RIOT_ENV_BASE_PATH``
+    Base directory under which virtual environments are created.
+    Defaults to ``.riot`` (relative to the current working directory).
+
+``RIOT_PIP_COMPILE_BACKEND``
+    Selects the compiler used when generating lockfiles via the
+    ``requirements`` command. Supported values:
+
+    - ``piptools`` (default) — invoke ``python -m piptools compile``;
```
brettlangdon, line 50: any reason to keep `piptools`? should we just switch riot to be `uv` based pip compile backend?

brettlangdon: then we can include `uv` as a project dependency for `riot` and don't need to worry about "is uv installed"

riot/venv.py
```diff
@@ -280,54 +280,125 @@ def __hash__(self):
...
+                )
+                sentinel.touch()
+            cmd = [
+                self.py.path(),
+                "-m",
+                "piptools",
+                "compile",
+                "-q",
+                "--no-annotate",
+                "--allow-unsafe",
```
brettlangdon, line 369: how does `uv` handle stuff like `setuptools`? we've been hit with that a ton o times, why we have added this here so we get `setuptools` locked in the requirements files as well

brettlangdon (commented):
suggestion, but in general lgtm

riot/venv.py
juanjux: It looks like `uv pip compile` includes setuptools, pip, wheel, and friends in the output by default, without needing a flag. So the practical effect is the same as `--allow-unsafe` on the pip-tools path: those packages still get pinned in `.riot/requirements/*.txt`.

docs/configuration.rst
juanjux: Because I am a coward? 😄  I am not sure if some user could, knowing it or not, be depending on some very specific format of the `.riot/requirements/*.txt` files. Probably not. Not sure what to do here...

brettlangdon: probably not since it is sort of a riot internal mechanism

as long as the existing requirements.txt files are still working/supported, and ones generated with uv are supported, then should be good.

I suspect they are forward/backwards compatible.

brettlangdon (approved):
I am ok with this as-is, if we find no issues during the transition, then migrating to be only uv compile based makes sense as a follow-up

e.g.

1. this PR
2. next PR adds `uv` as project dep and change default to `uv`
3. remove backend selection and only use `uv`


given we are primarily the only users, we can probably move a little quicker than that, but all in all seems good

